PR #4, opened 2020-11-25
Kinetics

ecisneros8:
This PR implements code generation for rate coefficients, reaction rates, and species production rates. Generated code is tested against Cantera. Rate coefficients are tested for a range of temperatures. Reaction rates and species production rates are tested for a set of autoignition states, as computed from Cantera. As for thermo, only the San Diego mechanism is used for testing. Other representative mechanisms will follow soon.

This PR completes code generation in that everything is in place to get chemical source terms for flow simulations for conventional combustion chemistry. Note that this PR does not implement MikeC's mirgecom-oriented changes.


pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        # Modified Arrhenius
+        return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
+
+
+def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
+    """This function returns an expression for the third-body rate factor"""
+    efficiencies = [react.efficiencies[sp] for sp in react.efficiencies]
+    indices_nondef = [sol.species_index(sp) for sp in react.efficiencies]
+    indices_default = np.delete(np.linspace(0, sol.n_species-1, sol.n_species,
+                                            dtype=int), indices_nondef)
```
inducer, line 207: Tip: Use `np.arange` instead of `np.linspace` for integers.

inducer, line 207:
```suggestion
    indices_default = [i for i in range(sol.n_species if i not in indices_nondef]
```
(easier to read maybe? but: higher complexity)

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+    elif b == 0.0 and a != 0.0 and t_a != 0.0:
+        # Classic Arrhenius rate
+        return p.Variable("exp")(np.log(a)-t_a/t)
+    else:
+        # Modified Arrhenius
+        return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
+
+
+def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
+    """This function returns an expression for the third-body rate factor"""
```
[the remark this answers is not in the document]
inducer: Particularly, document what *c* is (data type and meaning).

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+    efficiencies = [react.efficiencies[sp] for sp in react.efficiencies]
+    indices_nondef = [sol.species_index(sp) for sp in react.efficiencies]
+    indices_default = np.delete(np.linspace(0, sol.n_species-1, sol.n_species,
+                                            dtype=int), indices_nondef)
+    sum_nondef = sum(eff_i * c[index_i] for eff_i, index_i
+                     in zip(efficiencies, indices_nondef))
```
inducer, line 209:
```suggestion
    sum_nondef = efficiencies @ c[indices_nondef]
```

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+    sum_default = react.default_efficiency * sum(c[i] for i in indices_default)
```
inducer, line 210:
```suggestion
    sum_default = react.default_efficiency * c[i.indices_default].sum()
```

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+
+
+# {{{ Rates of progress
+
+def rate_of_progress_expr(sol: ct.Solution, react: ct.Reaction, c, k_fwd, k_eq):
+    """This function returns an expression for the reaction rate of progress"""
+    indices_reac = [sol.species_index(sp) for sp in react.reactants]
+    indices_prod = [sol.species_index(sp) for sp in react.products]
+
+    if len(react.orders) > 0:
```
inducer, line 235:
```suggestion
    if react.orders:
```
https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        nu_reac = [react.orders[sp] for sp in react.orders]
+    else:
+        nu_reac = [react.reactants[sp] for sp in react.reactants]
+
+    r_fwd = reduce(lambda x, y: x*y, [c[index]**nu for index, nu
```
inducer, line 240: > `lambda x, y: x*y`

https://docs.python.org/3/library/operator.html#operator.mul

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+                                      in zip(indices_reac, nu_reac)])
```
inducer, line 241: Use `np.prod`?

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+
+    if react.reversible:
+        nu_prod = [react.products[sp] for sp in react.products]
+        r_rev = reduce(lambda x, y: x*y, [c[index]**nu for index, nu
+                                          in zip(indices_prod, nu_prod)])
```
inducer, line 246: `np.prod`?

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        return k_fwd[int(react.ID)-1] * (r_fwd - k_eq[int(react.ID)-1] * r_rev)
```
inducer, line 247: For my education: what's the `int()` cast doing here?

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+    indices_fwd = [int(react.ID)-1 for react in sol.reactions()
+                   if species in react.reactants]
+    indices_rev = [int(react.ID)-1 for react in sol.reactions()
+                   if species in react.products]
+    nu_fwd = [sol.reactant_stoich_coeff(sol.species_index(species), react_index)
+              for react_index in indices_fwd]
+    nu_rev = [sol.product_stoich_coeff(sol.species_index(species), prod_index)
+              for prod_index in indices_rev]
+    sum_fwd = sum(nu*r_net[index] for nu, index in zip(nu_fwd, indices_fwd))
+    sum_rev = sum(nu*r_net[index] for nu, index in zip(nu_rev, indices_rev))
```
inducer, line 267: (Analogous to the above.)

pyrometheus/__init__.py
```diff
@@ -324,6 +410,76 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
+        falloff_center = np.log10(
+            ${cgm(troe_falloff_expr(react, Variable("T")))})
+        log_rp = np.log10(reduced_pressure[${i}])
+        c = -0.4-0.67*falloff_center
+        n = 0.75-1.27*falloff_center
+        f = (log_rp+c)/(n-0.14*(log_rp+c))
+        falloff_function[${i}] = 10.0**(falloff_center/(1.0+f**2))
+        %else:
+        falloff_function[${i}] = 1.0
+        %endif
```
inducer, line 434: I wish we could indent this somehow. But all the ways I can think of don't necessarily make it better. Have any ideas?

pyrometheus/__init__.py
```diff
@@ -324,6 +410,76 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
 
         return T
 
+    def get_falloff_rates(self, T, C, k_fwd):
+        k_high = np.zeros(self.num_falloff)
+        k_low = np.zeros(self.num_falloff)
```
inducer, line 415: It's better (read: easier to read, friendlier to lazy eval) to create new data structures rather than to make "empty" ones and then fill them with the desired values.

Imagine this code being run in a setting where `T` is an array of temperatures that lives on a GPU. :)

pyrometheus/__init__.py
```diff
@@ -342,6 +498,28 @@ def gen_python_code(sol: ct.Solution):
...
+        troe_falloff_expr=troe_falloff_expr,
+        rate_of_progress_expr=rate_of_progress_expr,
+        production_rate_expr=production_rate_expr,
+
+        # falloff_flags=[isinstance(r, ct.FalloffReaction)
+        #                for r in sol.reactions()],
+        # non_falloff_flags=[not isinstance(r, ct.FalloffReaction)
+        #                    for r in sol.reactions()],
+        # third_body_flags=[isinstance(r, ct.ThreeBodyReaction)
+        #                   for r in sol.reactions()],
```
inducer, line 512: ?

test/test_codegen.py
```diff
@@ -29,15 +29,82 @@
...
+        temp = reactor.T
+        rho = reactor.density
+        y = reactor.Y
+        # Prometheus kinetics
+        c = ptk.get_concentrations(rho, y)
+        r_pm = ptk.get_net_rates_of_progress(temp, c)
+        omega_pm = ptk.get_net_production_rates(rho, temp, y)
+        # Print
+        err_r = np.abs((r_ct-r_pm)/r_ct).max()
+        err_omega = np.abs((omega_ct[0:-1]-omega_pm[0:-1])/omega_ct[0:-1]).max()
```
inducer, line 70: Use `la.norm(..., np.inf)`.

test/test_codegen.py
```diff
@@ -29,15 +29,82 @@
...
+        # Compare
+        assert err_r < 1.0e-10
+        assert err_omega < 1.0e-8
```
inducer, line 73: What are the dominant parts of this error? Do we have any reason to believe that these tolerances would also work for different mechanisms?

inducer (commented):
Thanks for working on this. Some comments below.

pyrometheus/__init__.py
ecisneros8: I couldn't get this suggestion to work. I get:

TypeError: unsupported operand type(s) for @: 'list' and 'Subscript'

pyrometheus/__init__.py
ecisneros8: I couldn't get this to work. I get: 

AttributeError: 'Subscript' object has no attribute 'sum'

pyrometheus/__init__.py
ecisneros8: Noted

pyrometheus/__init__.py
ecisneros8: This is irrelevant now, but does this mean I should've used

`lambda x, y: mul(x,y)`

?

pyrometheus/__init__.py
ecisneros8: Changed, consistent with previous comment.

pyrometheus/__init__.py
ecisneros8: I need the reaction index but, to the best of my knowledge, there is no method in ct.Reaction that provides it. Instead, it has ID, which is a string, so I need to cast it to integer to use it as an index. Also note that ID is 1-based, hence the minus 1.

pyrometheus/__init__.py
ecisneros8: I no longer compute falloff rates this way, so this bit of code is gone. This change is due to your subsequent comment about empty initialization.

pyrometheus/__init__.py
ecisneros8: I've changed how I handle rate coefficients to (I think) follow this suggestion more closely.

pyrometheus/__init__.py
ecisneros8: Removed. Sorry!

pyrometheus/__init__.py
ecisneros8: Analogous to the previous comment.

test/test_codegen.py
ecisneros8: These tolerances were too loose and I can't recall why. One reason I suspect is I was taking relative errors of very small quantities. I've tightened them to 1e-12. This accuracy is enough to reproduce published ignition times. Also, it works for San Diego and UIUC, but I'll bring in larger mechs to test against.

test/test_codegen.py
ecisneros8: Noted. Changed where relevant.

pyrometheus/__init__.py
```diff
@@ -178,8 +179,93 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
 
 # {{{ Rate coefficients
 
-def arrhenius_expr(rate_coeff: ct.Arrhenius):
-    raise NotImplementedError()
+def rate_coefficient_expr(rate_coeff: ct.Arrhenius, t):
+    """This function returns rate-coefficient expressions"""
```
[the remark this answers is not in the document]
ecisneros8: I think I went through all of these, but if I missed one let me know.

test/test_codegen.py
MTCam: FWIW, both of these pass to 1e-10 for our current mechanisms. See e.g. same test in #7

pyrometheus/__init__.py
```diff
@@ -216,8 +225,111 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        # Classic Arrhenius rate
+        return p.Variable("exp")(np.log(a)-t_a/t)
+    else:
+        # Modified Arrhenius
+        return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
+
+
+def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
+    """:returns: The third-body concentration expression for reaction *react* in terms
+    of the species concentraions *c* as a :class:`pymbolic.primitives.Expression`"""
```
inducer, line 251: This isn't quite right, syntax-wise. I've fixed this up in #12. (OK to resolve once #12 is merged into this.)

inducer, line 251: What data type is *c*?

inducer, line 251:
```suggestion
    """:returns: The third-body concentration expression for reaction *react* in terms
    of the species concentrations *c* as a :class:`pymbolic.primitives.Expression`"""
```

pyrometheus/__init__.py
```diff
@@ -216,8 +225,111 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        nu_reac = [react.orders[sp] for sp in react.orders]
+    else:
+        nu_reac = [react.reactants[sp] for sp in react.reactants]
+
+    r_fwd = np.prod([c[index]**nu for index, nu in zip(indices_reac, nu_reac)])
+
+    if react.reversible:
+        nu_prod = [react.products[sp] for sp in react.products]
+        r_rev = np.prod([c[index]**nu for index, nu in zip(indices_prod, nu_prod)])
+        return k_fwd[int(react.ID)-1] * (r_fwd - k_eq[int(react.ID)-1] * r_rev)
```
inducer, line 309:
```suggestion
        # FIXME: It's not clear that this is available other than by this clunky,
        # string-parsing route
        reaction_index = int(react.ID)-1

        return k_fwd[reaction_index] * (r_fwd - k_eq[reaction_index] * r_rev)
```

pyrometheus/__init__.py
```diff
@@ -362,17 +474,102 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
+        falloff_function = np.array([
+        %for i, react in enumerate(falloff_reactions):
+            ${cgm(falloff_function_expr(
+                react, i, Variable("T"), Variable("reduced_pressure"),
+                Variable("falloff_center")))},
+        %endfor
+        ])*reduced_pressure/(1+reduced_pressure)
+
+        %for i, react in enumerate(falloff_reactions):
+        k_fwd[${int(react.ID)-1}] = k_high[${i}]*falloff_function[${i}]
```
inducer, line 516: Wouldn't `i` match `int(react.ID)-1`?

pyrometheus/__init__.py
```diff
@@ -362,17 +474,102 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
+            ${cgm(rate_coefficient_expr(react.rate, Variable("T")))},
+        %endif
+        %endfor
+        ])
+        %if falloff_reactions:
+        self.get_falloff_rates(T, C, k_fwd)
+        %endif
+
+        %for react in three_body_reactions:
+        k_fwd[${int(react.ID)-1}] *= (${cgm(third_body_efficiencies_expr(
```
inducer, line 535: Wouldn't the index of `react` match `int(react.ID)-1`?

inducer (commented):
A few comments below.

pyrometheus/__init__.py
ecisneros8: Not in this case. `int(react.ID-1)` is the reaction index in the full reaction set, whereas `i` is its index in the set of falloff reactions only. This is (maybe?) made more clear by noting that `k_fwd` is of size `num_reactions`, but `k_high` and `falloff_function` are of size `num_falloff`.

pyrometheus/__init__.py
ecisneros8: Analogous to the above, though for three-body reactions.

pyrometheus/__init__.py
inducer: OIC. Thanks for explaining!

pyrometheus/__init__.py
inducer: OIC. Thanks for explaining!

pyrometheus/__init__.py
ecisneros8: I don't see a difference in #12 . I'm looking at this line:

https://github.com/ecisneros8/pyrometheus/blob/610bb688c08a8b55b4c47b88d107dd1374eb0610/pyrometheus/__init__.py#L232

Am I lost?

ecisneros8: """:returns: A :class:`pymbolic.primitives.Expression` for the third-body concentration of reaction *react* in terms of a :class:`pymbolic.primitives.Expression` for the species concentrations *c*"""
?

MTCam: I have fixed this up in #7

pyrometheus/__init__.py
```diff
@@ -216,8 +225,111 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+        return a
+    elif b != 0 and t_a == 0:
+        # Weakly temperature-dependent rate
+        return a * t**b
+    elif b == 0 and a != 0 and t_a != 0:
+        # Classic Arrhenius rate
+        return p.Variable("exp")(np.log(a)-t_a/t)
+    else:
+        # Modified Arrhenius
+        return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
```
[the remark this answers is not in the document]
ecisneros8: @MTCam brought these unresolved conversations to my attention. These were addressed a long time ago, but I forgot to come back here and respond. Apologies. Here are some comments:
1. I wrapped `efficiencies` in `np.array()`.
2. I left `sum_default` as is. 
3. I got rid of unnecessary conditionals.

pyrometheus/__init__.py
ecisneros8: Resolved?

pyrometheus/__init__.py
MTCam: I'm worried about this in the MIRGE context. Will this still work properly when concentrations are multi-valued DOFArray type objects?

pyrometheus/__init__.py
MTCam: :thinking: worried about this one too a bit.

pyrometheus/__init__.py
MTCam: Yes, if @inducer had it fixed up in #12, then #7 resolves it.

MTCam: This one I'm not so sure about.  The species concentrations is a DOFArray in MIRGE-Com, and an np.array otherwise.  I wonder how we should document its type.  I don't think it would be right to mention DOFArray down inside Pyro.  Maybe we need to document `usr_np array type` or so... and then refer to it?  Just not sure.  I think this is something we can discuss - and (at least from my perspective) should not prevent merging.

MTCam (approved):
👍 This looks good to me, esp after #7 merge, ha :)

inducer:
FYI I'm doing a round of review here.

inducer:
Found it hard to review this while keeping #7 in mind. Instead, I've merged this into #7 changed its target branch to master.

PR #5, opened 2020-12-02
Param tests

MTCam:
* adds a simple option to write the code to file for human inspection
* parametrize the tests so that they hit all the available mechanisms

@inducer, @ecisneros8

### DIFF
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -562,6 +562,7 @@ def gen_python_code(sol: ct.Solution, file=None):
     exec_dict = {}
     exec(compile(code, "<generated code>", "exec"), exec_dict)
     exec_dict["_MODULE_SOURCE_CODE"] = code
+    exec_dict["_MODULE_SOURCE_CODE"] = code
     return exec_dict["Thermochemistry"]
 
 
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -525,7 +525,7 @@ class Thermochemistry:
 # }}}
 
 
-def gen_python_code(sol: ct.Solution, file=None):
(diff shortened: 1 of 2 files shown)


pyrometheus/__init__.py
```diff
@@ -555,7 +555,10 @@ def gen_python_code(sol: ct.Solution):
...
                                             for r in sol.reactions()])),
     )
-    print(code)
+    if file is not None:
+        print(code, file=file)
+    else:
+        print(code)
     exec_dict = {}
     exec(compile(code, "<generated code>", "exec"), exec_dict)
     return exec_dict["Thermochemistry"]
```
inducer, line 564: This is just factored wrong. There should be a function that returns the generated string.

pyrometheus/__init__.py
```diff
@@ -555,7 +555,10 @@ def gen_python_code(sol: ct.Solution):
                                            [isinstance(r, ct.ThreeBodyReaction)
                                             for r in sol.reactions()])),
     )
-    print(code)
+    if file is not None:
+        print(code, file=file)
+    else:
+        print(code)
     exec_dict = {}
     exec(compile(code, "<generated code>", "exec"), exec_dict)
```
inducer, line 563:
```suggestion
    exec(compile(code, "<generated code>", "exec"), exec_dict)
    exec_dict["_MODULE_SOURCE_CODE"] = code
```
Using this, you can see the generated code in https://documen.tician.de/pudb/.

MTCam: This was more about getting the actual py file than it was about putting eyes on the code.  It isn't essential.

MTCam:
I think this is what you had in mind @inducer?

inducer:
7e11e6a

PR #7, opened 2020-12-16
Kinetics mirge generator

MTCam:
Here's a set of changes to the Pyrometheus code generator that I will be testing in _MIRGE-Com_.   These pass the original _pyrometheus_ tests in the `kinetics` branch.   

This branch is required by the [_MIRGE-Com_ Pyrometheus EOS](https://github.com/illinois-ceesd/mirgecom/pull/236).


MTCam:
I am happy to report that [mirgecom EOS and Chem tests ](https://github.com/illinois-ceesd/mirgecom/pull/164)are passing with this version of the generator.  The `sanDiego` mechanism is failing due to the treatment of the falloff functions which are not present in UIUC mech.

ecisneros8:
@MTCam Did you branch from kinetics? I had falloff functions working (read: passing tests) there since quite a few commits ago. Look here: 

[](https://github.com/inducer/pyrometheus/blob/e3ff514f09e38db02b34da4b0b9cbd6530cb177c/pyrometheus/__init__.py#L235)

MTCam:
> @MTCam Did you branch from kinetics? I had falloff functions working (read: passing tests, for San Diego) there since quite a few commits ago. Look here:
> 
> [https://github.com/inducer/pyrometheus/blob/e3ff514f09e38db02b34da4b0b9cbd6530cb177c/pyrometheus/**init**.py#L235](falloff in kinetics branch)

Yes, the base branch of this PR is the kinetics branch in #4.  The generator in this branch still passes the native tests including `sanDiego` mechanism + falloff functions.  It is the MIRGE-specific multi-point test that is failing for `sanDiego` in MIRGE-Com only.

inducer:
I just merged master into #4. Please bring this up to date with respect to that branch, and add documentation. In addition, `npctx` is not a good name. Maybe `species_array_np` (or `species_np` or `spec_ary_np`) is a better name.

ecisneros8:
Thanks @inducer I’m starting documentation today.

pyrometheus/__init__.py
```diff
@@ -329,37 +330,84 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+            actx_np.log10(T)
+            actx_np.exp(T)
+            actx_np.where(T > 0, T_yes, T_no)
+            actx_np.linalg.norm(T, np.inf)
+            where T is model user data having the same size, shape, and nature
+            of the user's temperature data, for example. This parameter allows
+            the user to provide a numpy work-alike that is capable of processing
+            the user's data structures, or even invoking custom kernels for the
+            associated operations. This parameter defaults to *actual numpy*, so
+            unless it is needed by the user (e.g. for GPU processing).
```
inducer, line 387: Nobody sees this docstring. Please move this to the documentation of the `Thermochemistry` interface at the top of this file.

inducer, line 387: This is a good start to document what is expect of `actx_np`, but it needs to be pulled out into a separate interface and actually make functional use of RST markup.

pyrometheus/__init__.py
```diff
@@ -329,37 +330,84 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+
 
 class Thermochemistry:
-    model_name    = ${repr(sol.source)}
-    num_elements  = ${sol.n_elements}
-    num_species   = ${sol.n_species}
-    num_reactions = ${sol.n_reactions}
-    num_falloff   = ${
-        sum(1 if isinstance(r, ct.FalloffReaction) else 0
+    def __init__(self, actx_np=np):
```
inducer, line 370: I maintain that `actx_np` is not acceptable as a name. It suggests that the only thing this is usable for is meshmode's array context. That's not true. Instead, we need to describe what it is used for, and what is expected of it. The documentation you provide below begins to define that interface.

pyrometheus/__init__.py
```diff
@@ -409,7 +435,7 @@ def get_concentrations(self, rho, Y):
         conctest = self.iwts * rho * Y
         zero = 0 * conctest[0]
         for i, conc in enumerate(conctest):
```
ecisneros8, line 437: Where is conc used in this loop?

pyrometheus/__init__.py
```diff
@@ -409,7 +435,7 @@ def get_concentrations(self, rho, Y):
         conctest = self.iwts * rho * Y
```
ecisneros8, line 435: Why conctest? My issue with this variable name is the 'test' part. Can we call it concentrations or simply c?

pyrometheus/__init__.py
```diff
@@ -868,42 +523,43 @@ def get_temperature(self, enthalpy_or_energy, t_guess, y, do_energy=False):
...
             t_i += dt
-            if self.npctx.linalg.norm(dt, np.inf) < tol:
+            if _pyro_norm(self.usr_np, dt, np.inf) < tol:
                 break
 
         return t_i
 
     def get_falloff_rates(self, T, C, k_fwd):
-        k_high = make_obj_array([
+        ones = (1.0 + T) - T
```
ecisneros8, line 532: I've seen this is a couple of places now, and I'm just curious as to what it does? Is it a way of creating an array of 1's?

pyrometheus/__init__.py
```diff
@@ -952,8 +609,8 @@ def get_net_rates_of_progress(self, T, C):
     def get_net_production_rates(self, rho, T, Y):
         C = self.get_concentrations(rho, Y)
         r_net = self.get_net_rates_of_progress(T, C)
-        ones = r_net[0] + 1.0 - r_net[0]
-        return make_obj_array([
+        ones = (1.0 + r_net[0]) - r_net[0]
```
ecisneros8, line 612: Analogous to my previous comment.

test/test_codegen.py
```diff
@@ -30,6 +30,17 @@
 import pytest
 
 
+# Write out all the mechanisms for inspection
+@pytest.mark.parametrize("mechname", ["uiuc", "sanDiego"])
+def test_generate_mechfile(mechname):
+    """This function tests that pyrometheus-generated code."""
+    sol = ct.Solution(f"{mechname}.cti", "gas")
+    mfile = open(f"{mechname}.py", "w")
```
ecisneros8, line 38: I like src_file, py_file, or even mech_file instead. But I'll live.

test/test_codegen.py
```diff
@@ -127,7 +143,9 @@ def test_get_thermo_properties(mechname):
         keq_err = np.abs((keq_pm - keq_ct) / keq_ct).max()
```
ecisneros8, line 143: This propagated from kinetics, but my PR has a comment by @inducer to use `np.linalg.norm` instead.

pyrometheus/__init__.py
```diff
@@ -360,9 +358,37 @@ def _my_make_array(res_list):
...
-    def __init__(self, npctx=np):
-        self.npctx = npctx
+    def __init__(self, usr_np=np):
+        \"""Initialize the mechanism object.
+
+        Parameters
+        ----------
+        usr_np
+            numpy-like namespace providing at least the following functions:
+            usr_np.log(T)
```
ecisneros8, line 377: It is difficult to understand that this is not uniquely tied to the temperature. Can we either (i) state we're using T and temperature as an example up front; or (ii) generalize and then provide T and temperature as an example? (Hoping this is clear, otherwise let me know.)

pyrometheus/__init__.py
MTCam: It's 1's, but importantly respecting the size, shape, and nature (let's say) of *T*, whatever *T* may be.

pyrometheus/__init__.py
MTCam: We *may* use it instead of `concs[index]`.  More testing is needed to make sure this flies.   I had trouble with it, and switched back to using indexed data structure in debugging.  I suggest we ignore this for now.

test/test_codegen.py
MTCam: The thing I absolutely hate about the np.linalg norm is that it processes the entire vector and erases any component-specific information.   Typically, when doing the numerical debugging - I like to see the error in each component separately because it can help guide one to the problem(s).  So far I have been unsuccessful at arguing for code that allows component-by-component inspection; so changed.

pyrometheus/__init__.py
MTCam: My attempts to move this documentation were not entirely successful.  I could not find any examples of how to document class member methods in a global docstring.   `make html` complains about my section headings.  I kludged it together and it currently makes an acceptably rendered (if incomplete) document.

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
-        assert h_err < 1.0e-13
-        for i, r in enumerate(sol.reactions()):
-            if r.reversible:
-                assert keq_err[i] < 1.0e-13
+    # Homogeneous reactor to get test data
+    init_temperature = 1500.0
+    equiv_ratio = 1.0
+    ox_di_ratio = 0.21
+    stoich_ratio = 3.0  # 0.5
+    i_fu = sol.species_index(f"{fuel}")
```
inducer, line 210:
```suggestion
    i_fu = sol.species_index(fuel)
```

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
-        assert s_err < 1.0e-13
-        assert h_err < 1.0e-13
-        for i, r in enumerate(sol.reactions()):
-            if r.reversible:
-                assert keq_err[i] < 1.0e-13
+    # Homogeneous reactor to get test data
+    init_temperature = 1500.0
+    equiv_ratio = 1.0
+    ox_di_ratio = 0.21
+    stoich_ratio = 3.0  # 0.5
```
inducer, line 209: Previously, this depended on the fuel, but it no longer does. Why?

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
+    i_fu = sol.species_index(f"{fuel}")
+    i_ox = sol.species_index("O2")
+    i_di = sol.species_index("N2")
+    x = np.zeros(ptk.num_species)
+    x[i_fu] = (ox_di_ratio*equiv_ratio)/(stoich_ratio+ox_di_ratio*equiv_ratio)
+    x[i_ox] = stoich_ratio*x[i_fu]/equiv_ratio
+    x[i_di] = (1.0-ox_di_ratio)*x[i_ox]/ox_di_ratio
+    sol.TPX = init_temperature, ct.one_atm, x
```
inducer, line 217: Is this used anywhere?

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
+        # Get state from Cantera
+        temp = reactor.T
+        rho = reactor.density
+        y = np.where(reactor.Y > 0, reactor.Y, 0)
+        # Prometheus kinetics
+        c = ptk.get_concentrations(rho, y)
+        r_pm = ptk.get_net_rates_of_progress(temp, c)
+        omega_pm = ptk.get_net_production_rates(rho, temp, y)
+        # Print
+        err_r = np.linalg.norm((r_ct-r_pm), np.inf)
```
inducer, line 237:
```suggestion
        err_r = np.linalg.norm(r_ct-r_pm, np.inf)
```

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
+        err_omega = np.linalg.norm((omega_ct[0:-1]-omega_pm[0:-1]), np.inf)
```
inducer, line 238: Comment on why the last element of the array is being left out.

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
+    x[i_fu] = (ox_di_ratio*equiv_ratio)/(stoich_ratio+ox_di_ratio*equiv_ratio)
+    x[i_ox] = stoich_ratio*x[i_fu]/equiv_ratio
+    x[i_di] = (1.0-ox_di_ratio)*x[i_ox]/ox_di_ratio
+    sol.TPX = init_temperature, ct.one_atm, x
+    reactor = ct.IdealGasConstPressureReactor(sol)
+    sim = ct.ReactorNet([reactor])
+
+    time = 0.0
+    for step in range(50):
+        time += 1.0e-7  # 1.0e-6 for H2
```
inducer, line 223: I'm confused by the H2 comment here. Could you explain?

test/test_codegen.py
```diff
@@ -174,37 +188,65 @@ def test_get_temperature(mechname):
...
-def test_get_thermo_properties(mechname):
+@pytest.mark.parametrize("mechname,fuel", [("uiuc", "C2H4"),
+                                       ("sanDiego", "H2")])
+def test_kinetics(mechname, fuel):
     """This function tests that pyrometheus-generated code
-    computes thermodynamic properties c_p, s_R, h_RT, and k_eq
-    correctly by comparing against Cantera"""
-    # Create Cantera and pyrometheus objects
+    computes the right rates of progress for given temperature
+    and composition"""
```
inducer, line 196: Specify "right". (I know you didn't add this, but now is as good a time as any.)

inducer (commented):
Thanks for working on this. A few thoughts below.

test/test_codegen.py
MTCam: Just a note to self(s). Ethylene fuel needed 1e-7 to be stable.   1e-6 was sufficient for H2 fuel.

inducer:
> Update pyrometheus/__init__.py 

@MTCam Could you try and enter sensible commit messages?

test/test_codegen.py
MTCam: I don't understand what you are asking for.

test/test_codegen.py
MTCam: Yes, it is used on the very next line to initialize the reactor initial state.

pyrometheus/__init__.py
```diff
@@ -454,57 +514,51 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
-                T = T_i
+        ones = (1 + enthalpy_or_energy) - enthalpy_or_energy
+        t_i = t_guess * ones
+
+        for iter in range(num_iter):
+            f = enthalpy_or_energy - he_fun(t_i, y)
+            j = -pv_fun(t_i, y)
+            dt = -f / j
+            t_i += dt
+            if _pyro_norm(self.usr_np, dt, np.inf) < tol:
```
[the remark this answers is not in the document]
MTCam: There are a couple of things to consider that tie into this. An easy way to think about the root of the issue is that the output temperatures are sensitive to the initial guess passed to this routine.   The ideal situation is that we pass the actual/current temperature as the initial guess - we typically settle for a "reasonable" guess or for the temperature at the last time step.

If an initial guess is far off from the actual temperature, then this Newton iter can (a) take much more iters to converge to good temp, (b) converge to the wrong result or (c) fail to converge within num_iter. (a), (b) are the main things we want to avoid with restarting the DV (at least temperature for now).  Additionally, since the output temps are sensitive to input temps - we *need* to restart temperature in order to get deterministic results through restart - and this is important.

Unless we fix the temperature restart issues, we can expect that we will regularly encounter situation (c),  but because the starting guess and the actual temperature are not close.  In this case, we don't want to pass back NAN, we want to actually find the valid temperature, and so reverting to a more expensive procedure to find it would be a reasonable thing to do.   Or - until we get it fixed we can set num_iter to something insanely high.  Heh.

However if we *fix* the temperature restart issue and start actually using the current(ish) temperature as the initial guess, then if we could stop the code on encountering condition (c), it'd be terrific.  

Another potential fix is to roll a routine that provides a deterministic guess for a starting temperature based on the input state.   Then we'd only need to save state as we strongly desire, and we'd still get a deterministic temperature calc.   The big problem there is coming up with a sane temperature estimator routine; but it might be relatively easy to roll mechanism=specific estimators right into the generated code.

test/test_codegen.py
```diff
@@ -30,56 +30,15 @@
...
-        omega_pm = ptk.get_net_production_rates(rho, temp, y)
-        # Compare
-        assert np.linalg.norm(r_ct-r_pm, np.inf) < 1e-10
-        assert np.linalg.norm(omega_ct-omega_pm, np.inf) < 1e-10
-
-    return
+    mech_file = open(f"{mechname}.py", "w")
+    code = pyro.gen_thermochem_code(sol)
+    print(code, file=mech_file)
+    mech_file.close()
```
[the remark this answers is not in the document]
MTCam: Got it.

test/test_codegen.py
MTCam: I don't know.  Alternative answer: these are not the array elements you are looking for. :jedi-hand-wave:

pyrometheus/__init__.py
```diff
@@ -329,37 +349,64 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+
+    return result
+
+
+def _pyro_norm(usr_np, argument, normord):
+    # Wrap norm for scalars
+    from numbers import Number
+    if isinstance(argument, Number):
+        return np.abs(argument)
+    return usr_np.linalg.norm(argument, normord)
```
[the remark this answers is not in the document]
MTCam: Could be.  This was your suggested fix after we discussed it some time back.

pyrometheus/__init__.py
```diff
@@ -329,37 +349,64 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
-    num_falloff   = ${
-        sum(1 if isinstance(r, ct.FalloffReaction) else 0
+    def __init__(self, usr_np=np):
+        self.usr_np = usr_np
+        self.model_name = ${repr(sol.source)}
+        self.num_elements = ${sol.n_elements}
+        self.num_species = ${sol.n_species}
+        self.num_reactions = ${sol.n_reactions}
+        self.num_falloff   = ${
+            sum(1 if isinstance(r, ct.FalloffReaction) else 0
```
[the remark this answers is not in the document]
MTCam: Not sure I remember why I changed this.  I think it was when I added the argument for passed-in user numpy-like namespace.

test/test_codegen.py
MTCam: I took this to mean you wanted to clarify that "right" means.   In this case right = Cantera-predicted.

test/test_codegen.py
MTCam: This setting is of no consequence to the actual test.   We are testing against Cantera - so as long as the Cantera state is specified identically to the Pyrometheus state, the test is valid.   Regardless, I switched this back to a fuel-dependent stoichiometric ratio.

pyrometheus/__init__.py
```diff
@@ -384,40 +431,53 @@ def get_mix_molecular_weight(self, Y):
...
-        return self.gas_constant * T * np.sum(Y * h0_RT * self.iwts)
-
-    def get_mixture_internal_energy_mass(self, T, Y):
-        e0_RT = self.get_species_enthalpies_RT( T ) - 1.0
-        return self.gas_constant * T * np.sum(Y * e0_RT * self.iwts)
+        concs = self.iwts * rho * Y
+        zero = 0 * concs[0]
+        for i, conc in enumerate(concs):
+            concs[i] = self.usr_np.where(concs[i] > 0, concs[i], zero)
+        return concs
```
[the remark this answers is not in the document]
MTCam: That does not work as intended:
```py
E   ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

pyrometheus/__init__.py
```diff
@@ -384,40 +431,53 @@ def get_mix_molecular_weight(self, Y):
...
+
+    def get_mixture_specific_heat_cp_mass(self, temperature, massfractions):
+        cp0_r = self.get_species_specific_heats_R(temperature)
+        cpsum = sum([massfractions[i] * cp0_r[i] * self.iwts[i]
+                     for i in range(self.num_species)])
```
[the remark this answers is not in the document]
MTCam: This pattern appears several times, but for different properties.  Regardless, I refactored this a bit.  See how it grabs you now.

MTCam:
Scanning back through all the comments, I believe (most) everything has been addressed.  There is an open question about how to handle the specialized norm function `_pyro_norm`, but perhaps we can address that in an issue or separate PR?

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: I scanned through pyopencl.Array interface and it wasn't apparent to me how I would fix this inside there.  I will start a PR for it if you can help guide my effort just a bit to where I should start.

pyrometheus/__init__.py
MTCam: I took a better stab at fixing this up including the naming and documentation.  I am hoping this is closer to what you are looking for.

pyrometheus/__init__.py
```diff
@@ -319,6 +338,7 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
     """:returns: Species production rate for species *species* in terms of
     the net reaction rates of progress *r_net* as a
     :class:`pymbolic>primitives.Expression`"""
+    ones = (r_net[0] + 1.0) - r_net[0]
```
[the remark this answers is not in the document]
MTCam: Yeah, this is very hokey code.  I added a `_zeros_like` to try to clean it up a bit, but I suspect once you see what I did there that we'll need to talk about it a bit further.

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: Ah, so I still need to iterate through the array to do that?   If this expression is correct, then this "works" for `pyrometheus` proper - but unsure if it will continue working with the array context.

```
for (...):
   concs[i] = max(concs[i], 0)
```

test/test_codegen.py
MTCam: Just to clarify (sol.TPX = ) sets an internal state for the _Cantera_ `sol` object.  Subsequent uses of the `sol` object depend on this data.

pyrometheus/__init__.py
MTCam: Plz see below

MTCam:
A couple of things:
 - rolled in the flake8 CI for generated code from #12 
 - updated the script to generate and flake8 *all* available mechs 
 - moved the mech input files to test/mechs/{mechname}.cti
 - moved generated mech output to test/mechs/{mechname}_mech.py

It might be nicer to have the `mechs` directory with the CTI files live at top level instead of down in tests.   Other than that I am happy with how this is working.

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: @inducer added [this PR](https://github.com/inducer/meshmode/pull/100) to extend norm to numbers.  :tada:

Testing....

.ci/generate-test-mech.py
```diff
@@ -0,0 +1,10 @@
+import cantera as ct
+import pyrometheus as pyro
+import glob
+
+test_mechs = glob.glob("test/mechs/*.cti")
+for mech in test_mechs:
+    mechname = mech[11:-4]
+    with open(f"test/mechs/{mechname}_mech.py", "w") as outf:
+        code = pyro.gen_thermochem_code(ct.Solution(f"{mech}", "gas"))
+        print(code, file=outf)
```
inducer, line 10:
```suggestion
        print(pyro.gen_thermochem_code(ct.Solution(f"{mech}", "gas")), file=outf)
```

.ci/generate-test-mech.py
```diff
@@ -0,0 +1,10 @@
+import cantera as ct
+import pyrometheus as pyro
+import glob
+
+test_mechs = glob.glob("test/mechs/*.cti")
+for mech in test_mechs:
+    mechname = mech[11:-4]
+    with open(f"test/mechs/{mechname}_mech.py", "w") as outf:
```
inducer, line 8: The sphinx bits in #12 rely on a mechanism appearing at `pyrometheus.thermochem_example`. (But I realize there are multiple mechanisms here. Maybe emit the first one into `pyrometheus.thermochem_example`, too?

MTCam: Oh sorry, I did not see that.  I'll get that fixed up.

Longer term, it might be nice if we had some doc support for the general interface, and then a little blurb about each mechanism - maybe from embedded documentation, in which case we'd want to have all of them represented in actual py code?

MTCam: I refactored it just a bit with a function that generates a specified mech py file from a specified cti file. See the utils in `.ci` and `.github/workflows` yml file `doc` action for the generation of the required example.

MTCam:
> @MTCam Could you try and enter sensible commit messages?

Of course!  I intended to do that.  I didn't realize the bits that I wrote were showing up only in the dropdown until it was too late.

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: It turns out that the issue is that `numpy.linalg.norm(5, np.inf)` is not supported, so we need our wrapper after all.

pyrometheus/__init__.py
```diff
@@ -329,123 +328,228 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+        return 1/np.dot(self.iwts, mass_fractions)
+
+    def get_concentrations(self, rho, mass_fractions):
+        concs = self.iwts * rho * mass_fractions
+        zero = _pyro_zeros_like(concs[0])
+        for i, conc in enumerate(concs):
+            concs[i] = self.usr_np.where(concs[i] > 0, concs[i], zero)
+        return concs
+
+    def sum_over_species(self, mass_fractions, proparray):
```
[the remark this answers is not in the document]
MTCam: Just need the function to be a "verb".   Try this on for size. 
8dda1a7

pyrometheus/__init__.py
MTCam: 62615a7

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: Turns out, didn't work.  If this change is made, the code stops getting the right answers on all the points in the MIRGE-Com context as tested [here](https://github.com/illinois-ceesd/mirgecom/pull/147).

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: 6a644b6

.ci/generate-test-mech.py
MTCam: ab15da3, 68a9507

test/test_codegen.py
MTCam: 642cfe6

test/test_codegen.py
MTCam: 4b6c51b

test/test_codegen.py
MTCam: Fixed this up in (6ea96c8)

pyrometheus/__init__.py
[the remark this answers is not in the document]
inducer: > An easy way to think about the root of the issue is that the output temperatures are sensitive to the initial guess passed to this routine.

Just so I have this clear: What you're saying is that the equation being solved here doesn't suffice to fully define the temperature, i.e. the  initial guess matters? (If so, that seems strange to me.)

> Unless we fix the temperature restart issues, we can expect that we will regularly encounter situation (c), but because the starting guess and the actual temperature are not close.

Sure, Newton can fail. But wouldn't you rather know that it did than not?

A different aspect is what this dependency does to some of the interfaces we have in mirgecom. IIUC what you're saying, then you can't reasonably ask for temperatures without supplying a starting guess. That means the EOS interface in mirgecom needs to change.

(If you see this in time, could you add these discussions to the agenda for tomorrow?)

test/test_codegen.py
```diff
@@ -128,26 +88,84 @@ def test_get_pressure(mechname):
...
+
+        keq_pm1 = ptk.get_equilibrium_constants(t)
+        print(f"keq1 = {keq_pm1}")
+        keq_pm = 1.0 / np.exp(ptk.get_equilibrium_constants(t))
+        keq_ct = sol.equilibrium_constants
+
+        print(f"keq_pm = {keq_pm}")
+        print(f"keq_cnt = {keq_ct}")
+        print(f"temperature = {t}")
+        # Exclude check on irreversible reactions
```
inducer, line 146: Why? Maybe add some context on why that's difficult.

inducer (commented):
Sorry about taking forever on this. Here are some thoughts on this PR.

test/test_codegen.py
MTCam: Pyro doesn't "do" irreversible reactions.  Help me out @ecisneros8 ?

pyrometheus/__init__.py
```diff
@@ -327,123 +326,232 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+        return rho * rt / mmw
+
+    def get_mix_molecular_weight(self, mass_fractions):
+        return 1/np.dot(self.iwts, mass_fractions)
+
+    def get_concentrations(self, rho, mass_fractions):
+        concs = self.iwts * rho * mass_fractions
+        zero = _pyro_zeros_like(concs[0])
+        for i, conc in enumerate(concs):
+            concs[i] = self.usr_np.where(concs[i] > 0, concs[i], zero)
```
ecisneros8, line 484: I found that it is messing up Jacobians computed using JAX. All Jacobian entries associated with species with zero concentration come back zero, even when they shouldn't be. (Note, even when a mass fraction is zero, its corresponding Jacobian entries can be non-zero). I got rid of it, and I was able to verify the Jacobian of get_net_production_rates against finite differences and analytical expressions.

I'd much prefer these sorts of squashing to be done outside of pyro, and this is one example of why it's not best practice. Hope this is clear. Let me know otherwise.

pyrometheus/__init__.py
```diff
@@ -327,123 +326,228 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+    .. method:: get_mixture_specific_heat_cv_mass(self, T, Y)
+    .. method:: get_mixture_enthalpy_mass(self, T, Y)
+    .. method:: get_mixture_internal_energy_mass(self, T, Y)
+    .. method:: get_species_specific_heats_r(self, T)
+    .. method:: get_species_enthalpies_rt(self, T)
+    .. method:: get_species_entropies_r(self, T)
+    .. method:: get_species_gibbs_rt(self, T)
+    .. method:: get_equilibrium_constants(self, T)
+    .. method:: get_temperature(self, H_or_E, T_guess, Y, do_energy=False)
+    .. method:: __init__(self, usr_np=numpy)
```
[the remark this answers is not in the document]
MTCam: Fixed in (2b2c989)

pyrometheus/__init__.py
```diff
@@ -327,123 +326,228 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
+
+        Specify a user-defined NUMPY namespace as (*usr_np*) to the constructor
+        of a given mechanism thermochemistry class.
+
+        usr_np
```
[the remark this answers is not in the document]
MTCam: Fixed in (2b2c989).

pyrometheus/__init__.py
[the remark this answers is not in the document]
MTCam: meh.  for me usr_np is more descriptive.  what does bulk mean?

.ci/generate-example-mech.py
```diff
@@ -0,0 +1,11 @@
+import sys
+import pyrometheus as pyro
+
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
MTCam: Fixed in (ba8e442).

.ci/generate-example-mech.py
```diff
@@ -0,0 +1,11 @@
...
+    narg = len(sys.argv)
```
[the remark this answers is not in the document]
MTCam: Fixed in (ba8e442).

pyrometheus/__init__.py
MTCam: So you allow negative concentrations in Pyro?

test/test_codegen.py
ecisneros8: Pyro does do irreversible reactions. We even compute their equilibrium constants, even though they are meaningless and never factor into the computation of the rates, to maintain the equilibrium-constants array at size N_reactions. However, we don't compare them against Cantera because, well, they are meaningless.

pyrometheus/__init__.py
```diff
@@ -327,123 +326,228 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
-        mmw = self.get_mix_molecular_weight( Y )
-        RT  = self.gas_constant * T
-        return p * mmw / RT
 
-    def get_pressure(self, rho, T, Y):
-        mmw = self.get_mix_molecular_weight( Y )
-        RT  = self.gas_constant * T
-        return rho * RT / mmw
+def _pyro_norm(usr_np, argument, normord):
+    # Wrap norm for scalars
```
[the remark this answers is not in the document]
MTCam: This one is a bit better I think (ec25fa3).

test/test_codegen.py
MTCam: So my current plan would be to replace the comment with:
`# Exclude meaningless check on equilibrium constants for irreversible reactions`

pyrometheus/__init__.py
MTCam: Can we try this: (3fcdd54)?

inducer:
Heads up! I've changed the target branch here to master.

pyrometheus/__init__.py
```diff
@@ -183,17 +161,34 @@ def gen(c, t):
 
 # }}}
 
+# {{{ Data-handling helper
 
-# {{{ equilibrium constants
+def _zeros_like(argument):
+    return 0 * argument
```
inducer, line 167:
```suggestion
    # FIXME: This mishandles NaNs.
    return 0 * argument
```

pyrometheus/__init__.py
```diff
@@ -216,124 +211,353 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
-    def get_pressure(self, rho, T, Y):
-        mmw = self.get_mix_molecular_weight( Y )
-        RT  = self.gas_constant * T
-        return rho * RT / mmw
+def _pyro_norm(usr_np, argument, normord):
+    \"""This works around numpy.linalg norm not working with scalars.
 
-    def get_mix_molecular_weight(self, Y):
-        return 1/np.dot( self.iwts, Y )
+    If the argument is a regular ole number, it just uses regular numpy.abs,
```
inducer, line 371:
```suggestion
    If the argument is a regular ole number, it uses :func:`numpy.abs`,
```

pyrometheus/__init__.py
```diff
@@ -216,124 +211,353 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+    otherwise use a linalg norm from regular numpy, or from the user's numpy
+    clone if one was specified.
```
inducer, line 373:
```suggestion
    otherwise it uses ``usr_np.linalg.norm``.
```

pyrometheus/__init__.py
```diff
@@ -216,124 +211,353 @@ def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
...
+            - ``usr_np.log10(X)`` (like :data:`numpy.log10`)
+            - ``usr_np.exp(X)`` (like :data:`numpy.exp`)
+            - ``usr_np.where(X > 0, X_yes, X_no)`` (like :func:`numpy.where`)
+            - ``usr_np.linalg.norm(X, np.inf)`` (like :func:`numpy.linalg.norm`)
+
+            where the "bulk array type" is a type that offers arithmetic analogous
+            to :class:`numpy.ndarray` and is used to hold all types of (potentialy
+            volumetric) "bulk data", such as temperature, pressure, mass fractions,
+            etc. This parameter defaults to *actual numpy*, so it can be ignored
+            unless it is needed by the user (e.g. for GPU processing).
```
inducer, line 439:
```suggestion
            unless it is needed by the user (e.g. for purposes of 
            GPU processing or automatic differentiation).
```

pyrometheus/__init__.py
```diff
@@ -342,37 +566,119 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
-
-        for iter in range( 0, num_iter ):
-            F    -= he_fun( T_i, Y )
-            J    -= pv_fun( T_i, Y )
-            dT    = - F / J
-            T_i  += dT
-            if np.abs( dT ) < tol:
-                T = T_i
+        ones = _pyro_zeros_like(enthalpy_or_energy) + 1.0
+        t_i = t_guess * ones
```
inducer, line 570: What does this do?

pyrometheus/__init__.py
```diff
@@ -342,37 +566,119 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
                 break
-            F = H_or_E
-            J = 0.0
-
-        T = T_i
 
-        return T
-
-""", strict_undefined=True)
+        return t_i
```
inducer, line 580:
```suggestion
        raise RuntimeError("Temperature iteration failed to converge")
```

pyrometheus/__init__.py
```diff
@@ -342,37 +566,119 @@ def get_temperature(self, H_or_E, T_guess, Y, do_energy=False):
...
+        ones = _pyro_zeros_like(enthalpy_or_energy) + 1.0
+        t_i = t_guess * ones
+
+        for iter in range(num_iter):
+            f = enthalpy_or_energy - he_fun(t_i, y)
+            j = -pv_fun(t_i, y)
+            dt = -f / j
+            t_i += dt
+            if _pyro_norm(self.usr_np, dt, np.inf) < tol:
                 break
```
inducer, line 578:
```suggestion
                return t_i
```

test/test_codegen.py
inducer:
```suggestion
        # Exclude meaningless check on equilibrium constants for irreversible reactions
```

inducer (approved):
Just a few things, good to go subject to these.

pyrometheus/__init__.py
inducer: Agree with @ecisneros8, we should probably do this outside.

pyrometheus/__init__.py
MTCam: I intended to allow the call:
`get_temperature(.., tguess=300.0, ...)`,
where tguess is a single scalar value.

MTCam:
Seems like we are basically good-to-go for merging #7, #4 to main and address mass clipping as a separate issue? @ecisneros8, @inducer

ecisneros8:
I'm ok with merging, but we should not forget array calculations with Numpy in #20. I started testing this weekend but had to stop and don't know if I can come back to it until Friday. Thoughts or comments?

MTCam:
> I'm ok with merging, but we should not forget array calculations with Numpy in #20. I started testing this weekend but had to stop and don't know if I can come back to it until Friday. Thoughts or comments?

We won't forget.  #20 is a follow-on PR that extends Pyro to work with numpy arrays.  It will not go away when #7 is merged.

inducer:
I say let's merge this.

PR #12, opened 2021-01-10
Use generated code with Flake8 and Sphinx

inducer:
This makes direct use of the generated code to check Flake8 and generate docs using Sphinx. This is forked from #4. To avoid producing lots of conflicts with #7, I've so far held off from fixing the Flake8 failures in the generated code.

@ecisneros8 Do you agree that this is an OK idea?

If so, then I think the most straightforward merge path might be either:
- for @MTCam to merge this into #7 and merge it that way
- or for us to wait for #7 to go into #4 and merge it after
Thoughts?

### DIFF
--- a/.ci/generate-test-mech.py
+++ b/.ci/generate-test-mech.py
@@ -1,0 +1,4 @@
+import cantera as ct
+import pyrometheus as pyro
+with open("pyrometheus/thermochem_example.py", "w") as outf:
+    outf.write(pyro.gen_thermochem_code(ct.Solution("test/sanDiego.cti", "gas")))
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,7 @@ jobs:
                 curl -L -O -k https://gitlab.tiker.net/inducer/ci-support/raw/master/ci-support.sh
                 . ./ci-support.sh
                 build_py_project_in_conda_env
+                python -m pip install flake8 pep8-naming flake8-quotes
                 python .ci/generate-test-mech.py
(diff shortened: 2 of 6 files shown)


ecisneros8:
I agree this is a good idea. Either path seems fine to me right now, though I sense less work down the line if we merge now into Mike's branch.

MTCam:
Yeah, this is fantastic.   I lean more toward deferring this to dedicated effort, however.  Soon, we'll have undergrad workers who could easily tackle this, too.

MTCam:
> > only these linting issues pop up remain
> 
> Don't care much about unused vars in generated code. We could just `--extend-ignore=F841`.

Or we could figure out how to not add the falloff call when it is unneeded (any ideas @ecisneros8 ?).   I think for now it is fine either way - we can relax it now and bring it back later if we can.   

With your permission, then, I'll just roll flake8 stuff over onto the kinetics-merge-generator?

ecisneros8:
I just pushed a small change that skips generating get_falloff_rates for mechanisms with no falloff reactions. Hoping this is what you had in mine @MTCam.

MTCam:
> I just pushed a small change that skips generating get_falloff_rates for mechanisms with no falloff reactions. Hoping this is what you had in mine @MTCam.

Now `kinetics-mirge-generator` generates mech code that passes flake8 (with the line-length and quote exceptions). :tada:

inducer:
:tada: Nice!

MTCam:
All of the stuff in this PR has been rolled into #7, including the doc improvements.   OK to close this @inducer ?

inducer:
Yup.

PR #18, opened 2021-03-26
Lazarus: Add Jax Support for Autodifferentiation of Pyrometheus-Generated Mechanisms

cmikida2:
The minor changes here are designed to allow for Pyrometheus to carry out its operations using [Jax numpy](https://jax.readthedocs.io/en/latest/jax.numpy.html) (and subsequently allow for automatic differentiation of Pyrometheus code). While most of the changes here simply amount to propagating `usr_np` to some places it wasn't, there are a handful of additional changes associated with things like the fact that Jax arrays [don't support in-place updates](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#in-place-updates).

cc @inducer


cmikida2:
(Also, credit due to @alexfikl for doing a large chunk of the leg work helping me look into this)

pyrometheus/__init__.py
```diff
@@ -352,10 +350,15 @@ def _pyro_make_array(res_list):
     \"""
 
     from numbers import Number
-    all_numbers = all(isinstance(e, Number) for e in res_list)
+    all_numbers = all(
+        isinstance(e, Number)
+        or (isinstance(e, usr_np.ndarray) and e.shape == ())
+        for e in res_list)
```
MTCam, line 356: We want to be careful here. The interface documentation for the user's numpy clone does not indicate that it should sport the (.ndarray), (.dot), or (.zeros_like) members.  I'm not sure that the [numpy clone object](https://github.com/inducer/meshmode/blob/ac37ba5bc53223f40c195e25b23778cf6a004329/meshmode/array_context.py#L59) we actually use has these features.

inducer:
Any clue why CI didn't run on this PR?

cmikida2:
Great question, I have absolutely no idea. Could it because I'm trying to merge it into `kinetics-mirge-generator`? Not sure why that would be a problem, but no idea what else it would be.

MTCam:
> Any clue why CI didn't run on this PR?

:shrug: Might be a setting on main?

MTCam:
> Great question, I have absolutely no idea. Could it because I'm trying to merge it into `kinetics-mirge-generator`? Not sure why that would be a problem, but no idea what else it would be.

I should have peaked before posting; it doesn't.   If I could do a drive-by on meshmode it'd be dead easy (from the looks of it) to add `zeros_like` :).   The array context already provides that.

MTCam:
> Any clue why CI didn't run on this PR?

Is it because it is a PR from a fork? Forks didn't work for a bit on mirgecom CI, either. There is a setting that I think @matthiasdiener handled for illinois-ceesd.

cmikida2:
> Any clue why CI didn't run on this PR?

FWIW, the CI will fail when it's run, because jax (and jaxlib) will need to be added. But yeah, not sure. I looked through the settings on my fork to make sure it wasn't a switch I flipped, but I didn't see anything.

matthiasdiener:
> > Any clue why CI didn't run on this PR?
> 
> Is it because it is a PR from a fork? Forks didn't work for a bit on mirgecom CI, either. There is a setting that I think @matthiasdiener handled for illinois-ceesd.

Imho, the config in `.github/workflows/ci.yaml` looks fine. Can you paste the settings shown here: https://github.com/ecisneros8/pyrometheus/settings/actions ?

MTCam:
> > > Any clue why CI didn't run on this PR?
> > 
> > 
> > Is it because it is a PR from a fork? Forks didn't work for a bit on mirgecom CI, either. There is a setting that I think @matthiasdiener handled for illinois-ceesd.
> 
> Imho, the config in `.github/workflows/ci.yaml` looks fine. Can you paste the settings shown here: [ecisneros8/pyrometheus/settings/actions](https://github.com/ecisneros8/pyrometheus/settings/actions) ?

No I cannot :)  Can @ecisneros8?

cmikida2:
> > > Any clue why CI didn't run on this PR?
> > 
> > 
> > Is it because it is a PR from a fork? Forks didn't work for a bit on mirgecom CI, either. There is a setting that I think @matthiasdiener handled for illinois-ceesd.
> 
> Imho, the config in `.github/workflows/ci.yaml` looks fine. Can you paste the settings shown here: https://github.com/ecisneros8/pyrometheus/settings/actions ?

Just to be sure, I checked these same settings in my fork, and it's allowing all actions, but yeah not sure about here.

Screen Shot 2021-03-26 at 6 48 36 PM

pyrometheus/__init__.py
alexfikl: FWIW, this the extra `isinstance` check is needed because `np.exp(10)` returns a scalar, but `jnp.exp(10)` returns jax's `ndarray` lookalike. Not quite sure how to unify that.

matthiasdiener:
> Any clue why CI didn't run on this PR?

See here for the solution: https://github.com/ecisneros8/pyrometheus/pull/19

inducer:
@MTCam did you mean to close this? I'm fairly confused by that.

MTCam:
> @MTCam did you mean to close this? I'm fairly confused by that.

Not at all! This must have been automated somehow when I merged #7?

inducer:
@ecisneros8, could you reopen this? I'm not able to.

ecisneros8:
@inducer looking into it

inducer:
There's a reopen just below:
grafik

inducer:
Thanks!

inducer:
CI is actually running now. :tada:

FWIW, there was a merge conflict in `get_concentrations` in the code that cut off negative mass fractions. In accordance with the discussion in #21, I decided to simply delete that code, which resolved the conflict. :)

inducer:
@ecisneros8 I suspect what you mean by "reopen #7" was "restore the branch for #7", and what you mean by "delete #7" is "re-delete the branch for #7". Does that sound right?

cmikida2:
Hi all - the CI is failing because I need to add Jax to the conda build. I'll be fixing this shortly.

MTCam:
> CI is actually running now. 🎉
> 
> FWIW, there was a merge conflict in `get_concentrations` in the code that cut off negative mass fractions. In accordance with the discussion in #21, I decided to simply delete that code, which resolved the conflict. :)

I think that the current form of the `get_concentrations` in main works properly for main, MIRGE, and Jax.  We may want to aim for keeping it how it is until we have full resolution on the species limiting.  Without it, MIRGE will fail.

ecisneros8:
@inducer, yes, sounds right

ecisneros8:
@MTCam, @inducer 

Here's another option: Pyro could offer a `clip_mass_fractions` function. It can be called from MIRGE before `get_net_production_rates`.

MTCam:
> @MTCam, @inducer
> 
> Here's another option: Pyro could offer a `clip_mass_fractions` function. It can be called from MIRGE before `get_net_production_rates`.

I like that fine, but it does not address my main reluctance to dive into species limiting in MIRGE. The issue is that we call Pyro all over the place, with data from many different places.  The species limiting is very nicely centralized where it is right now - and requires no extra steps.  We _could_ put it in MIRGE at the very beginning of all of our operators - which would help centralize it (sort of).     

I think I would like it better if pyro had an option that would turn species clipping on or off for all of time instead of an interface to call.

While I agree 100% with the sentiment of wanting to make the caller do it, I feel that it places undue onus on the developer of the simulation capability that wants to use pyro in the EOS-type capacity.

inducer:
Uh, guys... I appreciate the discussion, but IMO it should be in #21, so as to not clutter up @cmikida2's PR (which is about something else entirely). I'm sure the check can be put back if and when we decide it should be.

inducer:
Uh-oh. This has picked up a conflict. @cmikida2, could you take a quick look?

cmikida2:
Will do!

.test-conda-env-py3.yml
```diff
@@ -7,3 +7,5 @@ dependencies:
 - python=3
 - cantera
 - pip
+- jax
+- jaxlib
```
inducer, line 11:
```suggestion
```
`jax` should already bring this in automatically: https://github.com/conda-forge/jax-feedstock/blob/7ebe55eae01e0ea909e27134842ee225a7e0326d/recipe/meta.yaml#L26

pyrometheus/__init__.py
inducer: > Maybe doing `getattr(e, 'shape', None) == ()` is enough?

Not loving our options here. I *might* be OK with this. As an alternate option, could we wrap `jnp` before passing it in as `usr_np` to incur the cost of this only with it?

pyrometheus/__init__.py
```diff
@@ -353,10 +352,15 @@ def _pyro_make_array(res_list):
...
```
inducer, line 358: For my education: What does Jax do without this special treatment?

inducer, line 358: This needs a comment on why it exists.

pyrometheus/__init__.py
```diff
@@ -375,7 +379,9 @@ def _pyro_norm(usr_np, argument, normord):
     # Wrap norm for scalars
     from numbers import Number
     if isinstance(argument, Number):
-        return np.abs(argument)
+        return usr_np.abs(argument)
+    if isinstance(argument, usr_np.ndarray) and argument.shape == ():
+        return usr_np.abs(argument)
```
inducer, line 384: Add a comment to explain its existence.

test/test_codegen.py
```diff
@@ -42,12 +46,13 @@ def test_generate_mechfile(mechname):
 
 
 @pytest.mark.parametrize("mechname", ["uiuc", "sanDiego"])
-def test_get_rate_coefficients(mechname):
+@pytest.mark.parametrize("np_instance", [np, jnp])
+def test_get_rate_coefficients(mechname, np_instance):
```
inducer, line 50: Can we make it so the tests dynamically detect whether jax is importable and adjust their set of numpys accordingly?

inducer, line 50:
```suggestion
@pytest.mark.parametrize("np_instance", [np, jnp])
def test_get_rate_coefficients(mechname, usr_np):
```
(for consistency)

inducer (commented):
Thanks!

pyrometheus/__init__.py
alexfikl: The entries in `res_list` are something like `jaxlib.DeviceArray(number, dtype=float32)` and if they're not recognized as scalars, they're put inside an object array in the next lines.

I don't recall if that messed up anything important or just at the end an `np.allclose` failed. I'm guessing it should have a similar effect as removing the special handling for numbers (?).

pyrometheus/__init__.py
alexfikl: Yeah, I also think wraping `jnp` a bit is the clearer way to go here. There are probably other differences that will pop up.

There's a `jax.device_get` that should get a nice number ouf of jax's `DeviceArray`, if necessary (pretty much the same as `pyopencl.array.Array.get`).

cmikida2: So apparently, indexing into a `jax.numpy` array full of scalars will give back an array of shape `()` rather than the number itself. I have been trying to create a wrapper for `jnp` but bearing that issue in mind, I'm not sure if this is something wrapping can fix...

test/test_codegen.py
cmikida2: Done by 8ccf61b.

pyrometheus/__init__.py
cmikida2: This is addressed differently by 8ccf61b, moving these functions that check for numbers into the generated class so they can be overwritten when Jax is in use, in a manner similar to https://github.com/ecisneros8/pyrometheus/pull/25.

test/test_codegen.py
cmikida2: Done by 8ccf61b.

pyrometheus/__init__.py
cmikida2: Done by 8ccf61b.

pyrometheus/__init__.py
cmikida2: Done by 8ccf61b.

.test-conda-env-py3.yml
cmikida2: Removed in 8e620cd.

pyrometheus/__init__.py
```diff
@@ -462,14 +425,57 @@ def __init__(self, usr_np=np):
         self.wts = ${str_np(sol.molecular_weights)}
         self.iwts = 1/self.wts
 
+    def pyro_make_array(self, res_list):
```
inducer, line 428: IMO this should stay underscored, so users don't get any ideas.

pyrometheus/__init__.py
```diff
@@ -339,46 +339,9 @@ def production_rate_expr(sol: ct.Solution, species, r_net):
...
-    otherwise it uses ``usr_np.linalg.norm``.
-    \"""
-    # Wrap norm for scalars
-    from numbers import Number
-    if isinstance(argument, Number):
-        return np.abs(argument)
-    return usr_np.linalg.norm(argument, normord)
-
-
 def _pyro_zeros_like(argument):
```
inducer, line 345: Move this, too, for consistency.

pyrometheus/__init__.py
```diff
@@ -462,14 +425,57 @@ def __init__(self, usr_np=np):
...
+        result = self.usr_np.empty((len(res_list),), dtype=object)
+
+        # 'result[:] = res_list' may look tempting, however:
+        # https://github.com/numpy/numpy/issues/16564
+        for idx in range(len(res_list)):
+            result[idx] = res_list[idx]
+
+        return result
+
+    def pyro_norm(self, argument, normord):
```
inducer, line 455: IMO this should stay underscored, so users don't get any ideas.

inducer (commented):
Thanks! This looks great. Just a few renames, then code-wise this is good to go IMO.

If we continue to see CI crashes, it might be worthwhile to isolate the Jax CI (and run it without pytest parallelism, if that helps).

inducer (approved):
Thanks!

PR #23, opened 2021-04-07
Troe Falloff Fix

ecisneros8:
Small bug in Troe-falloff expression that leads to problems in larger mechanisms that we don't test for. I found this while producing verification results for thesis. I can add test on a 32-species Ethylene mechanism to back this up if needed.

### DIFF
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -252,7 +252,7 @@ def troe_falloff_expr(react: ct.Reaction, t):
     troe_params = react.falloff.parameters
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    if troe_params[3] > 1.0e-16:
+    if np.abs(troe_params[3]) > 1.0e-16:
         troe_3 = p.Variable("exp")(-troe_params[3]/t)
         return troe_1 + troe_2 + troe_3
     else:
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -252,7 +252,8 @@ def troe_falloff_expr(react: ct.Reaction, t):
     troe_params = react.falloff.parameters
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    if np.abs(troe_params[3]) > 1.0e-16:
+    # if np.abs(troe_params[3]) > 1.0e-16:
+    if troe_params[3]:
         troe_3 = p.Variable("exp")(-troe_params[3]/t)
         return troe_1 + troe_2 + troe_3
     else:
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -252,7 +252,6 @@ def troe_falloff_expr(react: ct.Reaction, t):
     troe_params = react.falloff.parameters
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    # if np.abs(troe_params[3]) > 1.0e-16:
     if troe_params[3]:
         troe_3 = p.Variable("exp")(-troe_params[3]/t)
         return troe_1 + troe_2 + troe_3
--- a/test/mechs/UConn32.cti
+++ b/test/mechs/UConn32.cti
@@ -1,0 +1,1671 @@
+"""
+****************************************************************************
+
+   A 32-Species Skeletal Mechanism for C2H4-air
+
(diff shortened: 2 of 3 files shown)


MTCam (approved):
I think this is fine.  The argument to `np.abs` here will always be a number it looks like.

inducer:
> I can add test on a 32-species Ethylene mechanism to back this up if needed.

IMO this would be a very good thing.

A point of view about software that I've adopted for myself is that anything that's not tested should be considered broken already. (Or rephrased more positively: any property you think you value in software should have an automated test to back it up.) So, if you'd like Troe fall-off reactions to continue to work... :)

pyrometheus/__init__.py
```diff
@@ -252,7 +252,7 @@ def troe_falloff_expr(react: ct.Reaction, t):
     troe_params = react.falloff.parameters
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    if troe_params[3] > 1.0e-16:
+    if np.abs(troe_params[3]) > 1.0e-16:
```
inducer, line 255: I'll latch onto this, still knowing next to nothing about chemistry, but a bit about floating point. :) It seems that you're testing against machine epsilon. Machine epsilon is an estimate of relative error, using it on things that aren't ratios is likely inherently broken. (The only exception if the denominator in the ratio is O(1) by definition, in which case a comment should explain this.)

inducer, line 255:
```suggestion
    if np.abs(troe_params[3]) > np.finfo(troe_params_dtype).eps:
```
I don't know if we expect `troe_params` to be anything other than double precision values, but I nonetheless find this form preferable because it makes the intention clear.

ecisneros8: i can't recall why I simply didn't do `if np.abs(troe_params[3]):`, which would actually make sense, right?

ecisneros8: > > I can add test on a 32-species Ethylene mechanism to back this up if needed.
> 
> IMO this would be a very good thing.
> 
> A point of view about software that I've adopted for myself is that anything that's not tested should be considered broken already. (Or rephrased more positively: any property you think you value in software should have an automated test to back it up.) So, if you'd like Troe fall-off reactions to continue to work... :)

I will put in the test.

inducer: Sure, I'd have no objection to that. But you don't even need the `np.abs` then. `if troe_params[3]` should be all you need. Floats are "truthy" when non-zero.

https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false

MTCam: That leads me to ask about the original bug...

If the thing you were doing before `if thing > stuff` was a bug that was fixed by changing it to `if np.abs(thing) > stuff`  and now you consider replacing that with just `if thing`, I just can't imagine what the issue is/was.  What was the original issue?  :)

inducer: Probably `thing` was negative?

ecisneros8: troe_params[3] appears in some but not all Troe falloff reactions. If present, then the falloff expression is different than if not. Now, I didn't anticipate that it could be _negative_ in some instances where it actually shows up, so I was getting the wrong rate coefficients for those cases. 

In the San Diego mechanism, it only shows up as positive, so our tests didn't detect it. So, as @inducer said, our Troe falloff is *broken*.

MTCam: Now I'm really confused.  If negative is a problem, then how does changing it to `if thing` fix the original bug?  Also if it showed up in SD mech, and it was positive, then how/why was the original (`if thing > eps`) actually a bug?

If I'm just being dense (which is likely), then plz feel free to ignore me.

ecisneros8: `if troe_params[3]` simply checks that troe_params[3] is non-zero (I think, right @inducer). 

There are mechanisms in which some reactions have a negative `troe_params[3]`, but if we only check that `troe_params[3] > eps`, then it will ignore `troe_params[3] < 0` cases, and produce the wrong expression for Troe falloff.

MTCam: Yup, being dense, but thanks for not ignoring me :)

pyrometheus/__init__.py
```diff
@@ -252,7 +252,8 @@ def troe_falloff_expr(react: ct.Reaction, t):
     troe_params = react.falloff.parameters
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    if troe_params[3] > 1.0e-16:
+    # if np.abs(troe_params[3]) > 1.0e-16:
```
inducer, line 255:
```suggestion
```

inducer (approved):

inducer:
I think I fixed Flake8 here, which was the one remaining failure.

PR #25, opened 2021-05-13
Do not enforce positivity in get_concentrations

### LINKED ISSUE #22: messing up AD Jacobians computed using JAX.
I found (... that np.where ...) is messing up Jacobians computed using JAX. All Jacobian entries associated with species with zero concentration come back zero, even when they shouldn't be. 

(Note, even when a mass fraction is zero, its corresponding Jacobian entries can be non-zero). I got rid of it, and I was able to verify the Jacobian of get_net_production_rates against finite differences and analytical expressions.

I'd much prefer these sorts of squashing to be done outside of pyro, and this is one example of why it's not best practice. Hope this is clear. Let me know otherwise.

_Originally posted by @ecisneros8 in https://github.com/ecisneros8/pyrometheus/pull/7#discussion_r603462914_

inducer:
Closes #22.

The reason I think this is reasonable because it's easy to override after the fact:

```py
class Thermochem:
    def get_pressure(self, u):
        return self.get_concentrations(u)**1.4
    
    def get_concentrations(self, u):
        return u/1.5
    
def get_thermochem():
    return Thermochem
    
def do_stuff_with_pyro() :
    tchem_cls = get_thermochem()
    
    class MyThermochem(tchem_cls):
        def get_concentrations(self, u):
            return u/1.5 if u > 0 else 0
    
    tchem = MyThermochem()
    print(tchem.get_pressure(-5))

do_stuff_with_pyro()
```

cc @cmikida2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,4 +78,39 @@ jobs:
                 python .ci/generate-example-mech.py test/mechs/sanDiego.cti pyrometheus/thermochem_example.py
                 build_docs
 
+    downstream_tests:
+        strategy:
+            matrix:
+                downstream_project: [mirgecom]
+        name: Tests for downstream project ${{ matrix.downstream_project }}
+        runs-on: ubuntu-latest
+        steps:
+        -   uses: actions/checkout@v2
+        -   name: "Main Script"
+            env:
+                DOWNSTREAM_PROJECT: ${{ matrix.downstream_project }}
+            run: |
+                git clone "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                cd "$DOWNSTREAM_PROJECT"
+                echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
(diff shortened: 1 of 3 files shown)


MTCam:
Did 3fcdd54 not fix #22?

inducer:
Personally, I would much prefer to leave it up to the user whether they would like this truncation by overriding `get_concentrations` locally. The recipe I describe above would allow that, so to my mind this is preferable to making the choice on behalf of the user on the pyrometheus end.

inducer:
My sense is that @ecisneros8 felt the same way.

inducer:
@MTCam You OK with that approach?

MTCam (approved):
Sorry for the delay! Let's see what shakes loose in *MIRGE-Com*!

inducer:
Hmm, good point. I've added a downstream CI for mirgecom in 544d519.

MTCam:
> Hmm, good point. I've added a downstream CI for mirgecom in [544d519](https://github.com/ecisneros8/pyrometheus/commit/544d5190754534eb27ddb68e3aa14c49ab8fd523).

Ah, cool. I didn't even know we could do that.

PR #29, opened 2021-06-19
Make jnp references in testing code safe for non-Jax users

cmikida2:
Fixes test-breaking references to Jax numpy in testing code by only incorporating them when jax numpy is found (I tested this in a Jax-less environment (as well as a Jax-enabled environment) this time, so we should be good to go). Sorry!


inducer:
Thanks! I didn't quite like the way the logic is set up (too much repetition, odd importlib approach), so I gave this a shot myself, see #30.

inducer:
Superseded by #30.

PR #30, opened 2021-06-19
Improve jax detection logic in tests

inducer:


MTCam (approved):
Are we going to need an array context-enabled Jax?

inducer:
> Are we going to need an array context-enabled Jax?

Yep, that's the plan.

PR #31, opened 2021-06-29
Keep concentration > 0 in get_concentration

anderson2981:
chemistry and eos appear to run afoul when the concentration goes below zero

this change does nothing to "correct" the issue, that would require separate filter to drive things in the right, non-negative, direction. Best implemented in another part of the code, or driver.

For now, just keep pyrometheus from considering species mass fractions below zero, as this crashes the code.

### DIFF
--- a/pyrometheus/__init__.py
+++ b/pyrometheus/__init__.py
@@ -492,8 +492,17 @@ class Thermochemistry:
                 %endfor
                 )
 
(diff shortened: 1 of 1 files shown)


inducer:
See #25. This can be done in mirgecom by subclassing the `Thermochemistry` class.

cc @MTCam

PR #36, opened 2021-10-01
Refactor towards language-independence

inducer:

PR #46, opened 2022-04-20
Add generators for other languages

inducer:
This takes the "second half" of #36 and puts it on top of #45, which it contains. This also fixes attribution for those commits.

cc @ecisneros8 @MTCam


ecisneros8:
Multi-language has been merged via @henryleberre work.

PR #49, opened 2022-05-10
Add species dependent transport

tulioricci:
The new functions/routines are required to have a mixture + temperature dependence on viscosity, heat conductivity and species diffusivity in a (soon to be pushed) transport model for mirge-com. The quantities are evaluated using the polynomial fit within Cantera and are based on kinetic theory of gases.

### DIFF
--- a/pyrometheus/codegen/python.py
+++ b/pyrometheus/codegen/python.py
@@ -453,9 +453,10 @@ class Thermochemistry:
 
 
 def gen_thermochem_code(sol: ct.Solution) -> str:
-    """For the mechanism given by *sol*, return Python source code for a class conforming
-    to a module containing a class called ``Thermochemistry`` adhering to the
-    :class:`~pyrometheus.thermochem_example.Thermochemistry` interface.
+    """For the mechanism given by *sol*, return Python source code for a class
+    conforming to a module containing a class called ``Thermochemistry``
+    adhering to the :class:`~pyrometheus.thermochem_example.Thermochemistry`
+    interface.
     """
     return code_tpl.render(
         ct=ct,
--- a/test/test_codegen.py
+++ b/test/test_codegen.py
@@ -48,9 +48,10 @@ def make_jax_pyro_class(ptk_base_cls, usr_np):
     class PyroJaxNumpy(ptk_base_cls):
 
         def _pyro_make_array(self, res_list):
-            """This works around (e.g.) numpy.exp not working with object arrays of numpy
-            scalars. It defaults to making object arrays, however if an array
-            consists of all scalars, it makes a "plain old" :class:`numpy.ndarray`.
+            """This works around (e.g.) numpy.exp not working with object
+            arrays of :mod:`numpy` scalars. It defaults to making object arrays,
+            however if an array consists of all scalars, it makes a "plain old"
+            :class:`numpy.ndarray`.
 
             See ``this numpy bug <https://github.com/numpy/numpy/issues/18004>`__
             for more context.


test/test_codegen.py
```diff
@@ -175,6 +175,127 @@ def test_get_pressure(mechname, usr_np):
...
+        kappa_ct = sol.thermal_conductivity
+        diff_ct = sol.mix_diff_coeffs
+
+        # Pyrometheus transport
+        mu = ptk.get_mixture_viscosity(t, y)
+        kappa = ptk.get_mixture_thermal_conductivity(t, y)
+        diff = ptk.get_mixture_diffusivity(t,ct.one_atm,y)
+
+        err_mu = np.abs(mu - mu_ct)
+        assert err_mu < 1.0e-7
```
inducer, line 287: How come this (and other places) only achieve seven digit accuracy, given that, presumably, the same thing is being computed?

pyrometheus/chem_expr.py
```diff
@@ -164,8 +167,61 @@ def _zeros_like(argument):
 
 # }}}
 
-# {{{ Equilibrium constants
 
+# {{{ Transport polynomials & mixture rules
+
+def transport_polynomial_expr(c, n, t):
```
inducer, line 173: Many of your docstrings look quite off in terms of formatting. Please render the documentation via sphinx to make sure they look OK.

ecisneros8:
Where is this PR relative to the Transport branch that I started? I'm guessing it forked from it?

test/test_codegen.py
ecisneros8: Yes, this is odd. We come within 1e-13 or 1e-14 for thermal conductivity and viscosity, and mass diffusivity should be no different.

pyrometheus/chem_expr.py
MTCam: @tulioricci , fwiw the doc _build_ itself is OK.  The warning is due to not having the example mechanism available when it builds the doc.  This line:

https://github.com/tulioricci/pyrometheus/blob/a49204befbe2c22deadcba1e3e412d9bf40b114d/.github/workflows/ci.yml#L88

in the CI config builds that example before running the doc build, which is why CI does not encounter this issue.

pyrometheus/chem_expr.py
```diff
@@ -173,26 +173,26 @@ def _zeros_like(argument):
...
+            c[0]
+            + c[1] * p.Variable("log")(t)
+            + c[2] * p.Variable("log")(t) ** 2
+            + c[3] * p.Variable("log")(t) ** 3
+            + c[4] * p.Variable("log")(t) ** 4
         )**n
     )
 
 
 def wilke_mixture_rule_expr(sol: ct.Solution, sp, x, mu):
```
ecisneros8, line 191: Rename to `viscosity_mixture_rule_expr`, or `viscosity_mixture_rule_wilke_expr`.

pyrometheus/chem_expr.py
```diff
@@ -204,20 +204,20 @@ def wilke_mixture_rule_expr(sol: ct.Solution, sp, x, mu):
         8*(1 + (w[sp]/w[j]))
     ) for j in range(sol.n_species)])
 
-    
-def species_mixture_rule_expr(sol: ct.Solution, sp, mmw, x, Dij):
+
+def species_mixture_rule_expr(sol: ct.Solution, sp, mmw, x, d_ij):
```
ecisneros8, line 208: Rename to `species_mass_diff_mixture_rule_expr`. Also, I prefer `diff_ij`, than simply `d_ij`.

pyrometheus/chem_expr.py
```diff
@@ -204,20 +204,20 @@ def wilke_mixture_rule_expr(sol: ct.Solution, sp, x, mu):
...
     """Generate code for species mixture rule.
        See Robert Kee "Chemically Reacting Flow" book, chapter 12.
-    
-    :returns: Expression for the species diffusion mixture rule 
+
+    :returns: Expression for the species diffusion mixture rule
```
ecisneros8, line 212: Please change "species diffusion" to "species mass diffusion"

pyrometheus/chem_expr.py
```diff
@@ -204,20 +204,20 @@ def wilke_mixture_rule_expr(sol: ct.Solution, sp, x, mu):
...
         for species *sp* in terms of species mole fractions *w*
-        and binary diffusion *Dij* as a :class:`pymbolic.primitives.Expression`
-        
+        and binary diffusion *d_ij* as a :class:`pymbolic.primitives.Expression`
```
ecisneros8, line 214: `d_ij` is a binary diffusivity, not binary diffusion.

pyrometheus/codegen/python.py
```diff
@@ -295,51 +295,63 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
         return self._pyro_make_array([
-                % for ii in range(sol.n_species):        
+                % for ii in range(sol.n_species):
                 % for jj in range(sol.n_species):
-                ${cgm(ce.transport_polynomial_expr(sol.get_binary_diff_coeffs_polynomial(ii,jj), 1, Variable("temperature")))},
-                % endfor                
-                % endfor  
-                ]).reshape((self.num_species,self.num_species))  
+                ${cgm(ce.transport_polynomial_expr(
+                      sol.get_binary_diff_coeffs_polynomial(ii,jj), 1,
```
ecisneros8, line 338: Add space after commas, as so: `(ii, jj)`

test/test_codegen.py
```diff
@@ -0,0 +1,619 @@
...
+    sol = ct.Solution(f"mechs/{mechname}.cti", "gas")
+    ptk_base_cls = pyro.get_thermochem_class(sol)
+    ptk = make_jax_pyro_class(ptk_base_cls, usr_np)
+
+    # Loop over temperatures
+    ntemp = 29
+    temp = np.linspace(200.0, 3000.0, ntemp)
+    ii = -1
+    for t in temp:
+        ii = ii + 1
```
ecisneros8, line 262: What is this counter? I don't see it being used elsewhere.

pyrometheus/codegen/python.py
```diff
@@ -295,51 +295,63 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+                      Variable("temperature")))},
+                % endfor
+                % endfor
+                ]).reshape((self.num_species, self.num_species))
 
     def get_mixture_diffusivity(self, temperature, pressure, mass_fractions):
```
ecisneros8, line 344: It is unclear what this does. The name is not descriptive enough. A mixture has thermal, momentum, and mass diffusivities. It appears to return an array of (leading) dimension `num_species`, which implies to me to be the species mass diffusivities. Is this correct? If so, please rename to `get_species_mass_diffusivities`.

pyrometheus/chem_expr.py
tulioricci: Done. Renamed to "viscosity_mixture_rule_wilke_expr".

pyrometheus/chem_expr.py
tulioricci: Done.

pyrometheus/chem_expr.py
tulioricci: Done.

pyrometheus/chem_expr.py
tulioricci: Done.

pyrometheus/codegen/python.py
tulioricci: Done.

test/test_codegen.py
```diff
@@ -0,0 +1,619 @@
...
+            err_diff = np.abs(diff[i] - diff_ct[i])
+            assert err_diff < 1.0e-13
+
+    return
+
+
+@pytest.mark.parametrize("mechname, fuel, stoich_ratio, dt", [("uiuc", "C2H4", 3.0,
+                                                               1e-7)])
+@pytest.mark.parametrize("usr_np", numpy_list)
+def test_get_transport_properties_temp(mechname, fuel, stoich_ratio, dt, usr_np):
```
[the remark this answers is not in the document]
tulioricci: Changed the name to "test_get_transport_properties_temperature" and the other test to "test_get_transport_properties_reactor".

test/test_codegen.py
```diff
@@ -0,0 +1,619 @@
...
+    """This function tests that pyrometheus-generated code
+    computes transport properties (viscosity and thermal conductivity)
```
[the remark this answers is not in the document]
tulioricci: Done.

test/test_codegen.py
tulioricci: I forgot to clean it up. Thanks.

test/test_codegen.py
```diff
@@ -0,0 +1,619 @@
...
+        i_di = sol.species_index("N2")
+
+        x = np.zeros(ptk.num_species)
+        x[i_fu] = (ox_di_ratio*equiv_ratio)/(stoich_ratio+ox_di_ratio*equiv_ratio)
+        x[i_ox] = stoich_ratio*x[i_fu]/equiv_ratio
+        x[i_di] = (1.0-ox_di_ratio)*x[i_ox]/ox_di_ratio
+
+        sol.TPX = t, ct.one_atm, x
+        sol.equilibrate("TP")
+        _, _, y = sol.TPY
```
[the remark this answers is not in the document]
tulioricci: Way better. Thanks.

pyrometheus/codegen/python.py
tulioricci: Thanks for pointing that out. Done.

test/test_codegen.py
```diff
@@ -0,0 +1,619 @@
...
+
+        mu_ct = sol.viscosity
+        kappa_ct = sol.thermal_conductivity
+        diff_ct = sol.mix_diff_coeffs
+
+        # Pyrometheus transport
+        mu = ptk.get_mixture_viscosity(t, y)
+        kappa = ptk.get_mixture_thermal_conductivity(t, y)
+        diff = ptk.get_mixture_diffusivity(t, ct.one_atm, y)
```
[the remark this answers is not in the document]
tulioricci: Done.

test/test_codegen.py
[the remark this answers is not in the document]
tulioricci: I split the previous test in two. Now the temperature test exercise both individual species and mixtures. The mixture test is necessary for the species mass diff. since it is also a function of the mixture, so it can not be evaluated for a single species only. 

Let me know your thoughts on this.

pyrometheus/codegen/python.py
```diff
@@ -295,51 +295,63 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
         return self._pyro_make_array([
                 % for sp in range(sol.n_species):
-                ${cgm(ce.transport_polynomial_expr(sol.get_viscosity_polynomial(sp), 2, Variable("temperature")))},
+                ${cgm(ce.transport_polynomial_expr(sol.get_viscosity_polynomial(sp),
+                                                    2, Variable("temperature")))},
                 % endfor
                 ])
-                
+
     def get_mixture_viscosity(self, temperature, mass_fractions):
```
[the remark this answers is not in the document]
tulioricci: I was following Cantera's nomenclature to use "mixture" for the rule (in opposition to multi-component) but I agree this creates confusion. I suggest changing this (and the other ones as well) to "get_mixture_averaged_viscosity" to be clear.

Let me know your thoughts on this.

pyrometheus/codegen/python.py
```diff
@@ -295,51 +295,63 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+        conductivities = self.usr_np.sqrt(temperature)\
+ * self.get_species_thermal_conductivities(temperature)
+        return 0.5*(sum(mole_fractions*conductivities)
             + 1/sum(mole_fractions/conductivities))
-                
+
     def get_species_diffusivities(self, temperature):
         return self._pyro_make_array([
-                % for ii in range(sol.n_species):        
+                % for ii in range(sol.n_species):
```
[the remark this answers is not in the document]
tulioricci: I tried but it failed miserably. Since you have more experience with Mako, you could help me out.

tulioricci (commented):
@ecisneros8 , implemented your suggestions and comments. Thanks for your inputs!

pyrometheus/codegen/python.py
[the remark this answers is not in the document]
ecisneros8: Ok I'll give it a shot.

test/test_codegen.py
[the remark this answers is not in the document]
ecisneros8: Why have two tests if a single one can accomplish the same goal?

pyrometheus/codegen/python.py
```diff
@@ -293,63 +293,62 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
         mole_fractions = self.iwts * mass_fractions\
  * self.get_mix_molecular_weight(mass_fractions)
-        conductivities = self.usr_np.sqrt(temperature)\
- * self.get_species_thermal_conductivities(temperature)
+        conductivities = self.get_species_thermal_conductivities(temperature)
         return 0.5*(sum(mole_fractions*conductivities)
             + 1/sum(mole_fractions/conductivities))
 
-    def get_species_diffusivities(self, temperature):
+    def get_species_mass_diffusivities(self, temperature):
```
ecisneros8, line 331: I think these are binary diffusivities.

pyrometheus/codegen/python.py
```diff
@@ -293,63 +293,62 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
                 ${cgm(ce.transport_polynomial_expr(
-                      sol.get_binary_diff_coeffs_polynomial(ii,jj), 1,
+                      sol.get_binary_diff_coeffs_polynomial(i, j), 1,
                       Variable("temperature")))},
                 % endfor
                 % endfor
                 ]).reshape((self.num_species, self.num_species))
 
-    def get_mixture_diffusivity(self, temperature, pressure, mass_fractions):
+    def get_mixture_averaged_mass_diffusivity(self,
```
ecisneros8, line 342: This function uses binary diffusivities to compute the diffusivity of _each_ species into a mixture of all others (I think). Thus, the more appropriate name would be `get_species_mixavg_diffusivities`. Note plural diffusivities-- this is important to denote we are returning an array of (leading) dimension `num_species`.

pyrometheus/codegen/python.py
ecisneros8: I left a couple of new comments regarding this, but I'll summarize here. We have binary diffusivities (how pairs of species diffuse into each other), and mixture-averaged diffusivities (how one species diffuses into a mixture of all others). It is important that our function names reflect this. Please be sure to use `diffusivities` in function names.

ecisneros8 (changes requested):
@tulioricci -- looking good. Please see these new comments. They tie back to open requests in the previous review. Thanks again for taking time to finish this!

pyrometheus/codegen/python.py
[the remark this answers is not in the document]
tulioricci: Thanks, let me know if you succeed. And sorry, before I meant to ask you the question "Could you help me out?" and not to force you to help me.

pyrometheus/codegen/python.py
tulioricci: Renamed to "get_species_binary_mass_diffusivities".

pyrometheus/codegen/python.py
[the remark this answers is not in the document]
tulioricci: Changed all the mixture averaged functions to:

get_mixture_viscosity_mixavg
get_mixture_thermal_conductivity_mixavg
get_mixture_mass_diffusivities_mixavg

pyrometheus/codegen/python.py
tulioricci: I renamed to "get_species_binary_mass_diffusivities" to agree with a comment below as well.

tulioricci:
@ecisneros8 , I added a function to return the diagonal of the binary species diffusivities. This is necessary for regions where only a single species is present (Yi -> 1.0) and, to keep consistency with Cantera, only the self-diffusivity is used instead of a mixture rule that becomes singular.

I prefer this way because it reduces the burden on mirge-com by avoiding the entire matrix, where some cases can go to millions of grid points.

ecisneros8:
@tulioricci I have a few questions about this approach:
- What does Cantera do in such instances? I don't recall it having separate functions to handle these cases.
- Can the mixture rule be reformulated to avoid the singularity? I vaguely remember working on these mixture rules and changing the denominator from say, `1 - X_i` to `1 + ...`. It entailed changing from mole to mass fractions (or vice versa?). We can chat more about this.
- Where should it be decided what function to call: in mirgecom or pyro? And why?

I think a more sensible approach is to have one function that checks for these cases and returns just the diagonal, else the full matrix.

tulioricci:
@ecisneros8 , I combined the two functions into a single one as you requested. The check for self-diffusivity was inside the code-generation file. Cantera does the selection of which diffusivity (i.e., mixture-averaged or self-diffusivity) for sum(Y, i != j) == 0. The checking is done inside the mixture-rule evaluation. I don't think we can replicate exactly Cantera here since we are working with symbolic variables and the mass fraction is not known during the code generation.

Here, I am adding an optional argument that can be passed by the user to control when the mixture-rule switches to self-diffusivity. This increases robustness in case Mirge-COM misbehaves during the time integration. Such parameter can not be exactly 0.0 or it will return NaN. However, machine-precision zero (1.0e-15) works fine and returns the expected values.

pyrometheus/codegen/python.py
```diff
@@ -286,6 +293,69 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
         emix = self.get_mass_average_property(mass_fractions, e0_rt)
         return self.gas_constant * temperature * emix
 
+    def get_species_viscosities(self, temperature):
```
inducer, line 296: Document these.

pyrometheus/codegen/python.py
```diff
@@ -286,6 +293,69 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+            + 1/sum(mole_fractions/conductivities))
+
+    def get_species_binary_mass_diffusivities(self, temperature):
+        return self._pyro_make_array([
+                % for i, j, in product(range(sol.n_species), range(sol.n_species)):
+                ${cgm(ce.transport_polynomial_expr(
+                      sol.get_binary_diff_coeffs_polynomial(i, j), 1,
+                      Variable("temperature")))},
+                % endfor
+                ]).reshape((self.num_species, self.num_species))
```
inducer, line 339: Will this shape work in both the array-of-values-on-grid (via object arrays) and the 0D reactor (all species in one numpy array) case?

pyrometheus/chem_expr.py
```diff
@@ -164,8 +167,63 @@ def _zeros_like(argument):
...
+            + c[3] * p.Variable("log")(t) ** 3
+            + c[4] * p.Variable("log")(t) ** 4
+        )**n
+    )
+
+
+def viscosity_mixture_rule_wilke_expr(sol: ct.Solution, sp, x, mu):
+    """Generate code for species mixture rule.
+       See Robert J. Kee, Michael E. Coltrin and Peter Glarborg
+       "Chemically Reacting Flow" book, chapter 12.
```
inducer, line 195: Fix indentation. (Should line up with rest of code.)

pyrometheus/chem_expr.py
```diff
@@ -164,8 +167,63 @@ def _zeros_like(argument):
...
+        1 + sqrt((mu[sp]/mu[j])*sqrt(w[j]/w[sp]))
+    )**2 / sqrt(
+        8*(1 + (w[sp]/w[j]))
+    ) for j in range(sol.n_species)])
+
+
+def species_mass_diff_mixture_rule_expr(sol: ct.Solution, sp, mmw, x, diff_ij):
+    """Generate code for species mixture rule.
+       See Robert J. Kee, Michael E. Coltrin and Peter Glarborg
+       "Chemically Reacting Flow" book, chapter 12.
```
inducer, line 213: Fix indentation. (Should line up with rest of code.)

inducer, line 213: Make this a proper reST citation.

pyrometheus/chem_expr.py
inducer, line 195: Make this a proper reST citation.

pyrometheus/codegen/python.py
```diff
@@ -286,6 +293,69 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+        mole_fractions = self.iwts * mass_fractions * mmw
+        diff_ij = self.get_species_binary_mass_diffusivities(temperature)
+        mix_rule_f = self._pyro_make_array([
+              % for sp in range(sol.n_species):
+              self.usr_np.where(
+                  self.usr_np.greater(1.0-mole_fractions[${sp}], threshold),
+                  ${cgm(ce.species_mass_diff_mixture_rule_expr(sol, sp,
+                      Variable("mmw"), Variable("mole_fractions"),
+                      Variable("diff_ij")))},
+                  diff_ij[${cgm(sp)}, ${cgm(sp)}],
```
inducer, line 353: This seems... inelegant. Is there a better way to do this?

inducer (commented):
This is from @ecisneros8 and me scrolling through this for a bit and chatting.

pyrometheus/codegen/python.py
tulioricci: @ecisneros8 , based on your experience with grids, what do you think?

pyrometheus/codegen/python.py
tulioricci: The "where" is necessary to switch from a mixture-rule to the self-diffusivity and avoid a singularity of 0/0. Tests with Mirge-com result in NaN in the very first iteration when not considering the self-diffusivity. In my opinion, it is better to check here which coefficient to use, instead of returning both D_i^m (mixture) and D_ii (self) to MIRGE-COM and do the checking over there.

If the comparison is done against a hard 1.0 (for instance "self.usr_np.less(mole_fractions[${sp}], 1.0)"), without considering round-off errors, it is possible that the code fails. For instance, the output of "print(dii_pm[i], dii_ct[i], y)", which is line 217 in the test_codegen.py results in:

5.092329409340231e-06 5.092329409340232e-06 [1. 0. 0. 0. 0. 0. 0.]
9.923190033460186e-06 9.923190033460188e-06 [0. 1. 0. 0. 0. 0. 0.]
4.869780035934058e-06 4.869780035934059e-06 [0. 0. 1. 0. 0. 0. 0.]
9.842024372037339e-06 9.842024372037342e-06 [0. 0. 0. 1. 0. 0. 0.]
8.234238345402362e-06 8.234238345402364e-06 [0. 0. 0. 0. 1. 0. 0.]
  7.37190664636879e-05 7.371906646368791e-05 [0. 0. 0. 0. 0. 1. 0.]
                                  nan 1.002002531371639e-05 [0. 0. 0. 0. 0. 0. 1.]
                                  
However, considering even a round-off error of 1e-15 fixed the problem and the code doesn't fail/crash. I added the "threshold" option to give the user the decision to consider when to switch from a mixture-rule to self diffusivity. This could be passed down from the final application (Mirge-COM, in my case) towards this function.

Please, let me know your thoughts on this.

pyrometheus/codegen/python.py
tulioricci: Added the theoretical formulation  and description in model.rst.

pyrometheus/chem_expr.py
tulioricci: Removed these two lines and added the proper citation.

pyrometheus/chem_expr.py
tulioricci: Removed these two lines and added the proper citation.

pyrometheus/chem_expr.py
tulioricci: Done.

pyrometheus/chem_expr.py
tulioricci: Done.

pyrometheus/codegen/python.py
tulioricci: The 0D reactor is being exercised with the new test functions. Now, if you are referring to grids as the Mirge-COM case, for instance, it is working with the current implementation. A PR is pending (https://github.com/illinois-ceesd/mirgecom/pull/715) where it exercises this capability (plus, I've been using this code in different cases over the last months).

PR #51, opened 2022-06-02
Upgrade to Cantera 2.6 with yaml input files

ecisneros8:
This PR upgrades Pyro to use Cantera's preferred input-file format-- yaml. The template and EGFs were lightly edited because:
- `react.ID` is no longer populated when the mech is read from a yaml file;
- `reaction.falloff` is no longer supported, now folded into `reaction.rate`.


MTCam (commented):
Thanks, Esteban.  This is great.  Did you already make a `uiuc.yaml`?  I would like to test mirgecom with this development but I don't see the yaml files yet.

ecisneros8:
@MTCam Yes, uiuc.yaml is there, as well as all other mechs we've been using for tests (San Diego UConn-32, GRI-3.0).

ecisneros8:
Tests are failing with this Cantera error:

`AttributeError: New implementation uses 'FalloffRate' objects that are retrieved using the 'rate' property.`

This is something in Cantera 2.6 with yaml inputs. I'm trying to reproduce this locally to address it.

ecisneros8:
@MTCam @inducer This is failing due to something downstream in mirgecom. I can't look into that.

MTCam:
> @MTCam @inducer This is failing due to something downstream in mirgecom. I can't look into that.

mirgecom needs to get updated to yaml files as well. another thing we could do, which might be better, is to import the config files directly from pyrometheus instead of locally from mirgecom.  We currently have an infrastructure that imports mirgecom-local CTI files.  If instead this was just something like `from pyrometheus import uiuc as Thermochem`, that'd be just terrific imo.  Then mirgecom could get out of the business of having local mech config files and whether they are YAML or CTI would be none of our business.

inducer:
> `from pyrometheus import uiuc as Thermochem`

I'm not a chemistry person, but this seems to be based on the idea that there either
- pyrometheus is a special-purpose tool that only exists to server mirgecom, or
- are only a finite number of mechanisms that could be of interest, and it is therefore appropriate for pyro to attempt to enumerate them.

As far as I can tell, both of these are false.

ecisneros8:
@inducer, sorry I messed up your previous comment...
But, in the spirit of your first bullet, we shouldn't be downstream testing mirgecom for CI or PRs. Unless I'm missing something.

inducer:
> But, in the spirit of your first bullet, we shouldn't be downstream testing mirgecom for CI or PRs. Unless I'm missing something.

I was making a distinct point there. While I believe it to be true that the goal for pyro lies beyond being a utility for mirgecom, I strongly feel that no library should needlessly break compatibility with an established interface. That is, any code that worked previously should just keep working. Interfaces that are no longer favored should be deprecated (a specific procedure for phasing them out) for a while before being removed.

The fact that these changes break mirgecom is a strong indication that these changes violate this rule. Python offers enough ways to maintain compatibility IMO that breakage is almost always avoidable.

inducer:
> but it has been in-use now for quite some time and has "spread" a bit.

I'll acknowledge that I know nothing about this community. That said, if Cantera doesn't see a need to ship mechanisms, let alone this "uiuc" one, why should pyro? The standard here, to me, should be, "is this of global relevance?" If there's even a shade of doubt, then I find that it has no business clogging up the pyro distribution.

> provide pre-generated classes

I still do not understand the technical reason for pre-generation. But I do understand its substantial technical disadvantages. Care to elaborate?

MTCam:
| That said, if Cantera doesn't see a need to ship mechanisms, let alone this "uiuc" one, why should pyro? 

I take your point, but @ecisneros8 is the _source_ of the `uiuc` mechanism; we (CEESD), are the consumers.   Since a correct "mechanism configuration" file is required in order to create a pyrometheus mechanism class, I think it makes sense for pyrometheus to provide one or more configurations for common mechanisms.   Another potential advantage is that we could be sure that the "mech config" that we were getting was up-to-date with the latest as it is developed and consistent with pyro@main.   Esteban and co-workers may update these mechanisms, and pyro along with it.

| The standard here, to me, should be, "is this of global relevance?" If there's even a shade of doubt, then I find that it has no business clogging up the pyro distribution.

But Pyro already includes all of these config files, whether they are CTI or YAML. I just want a (light) interface to get them.

A concern I have with our (CEESD) just copying the CTI or YAML file as-is, is similar to the concern that our pre-generated mechs fall out-of-sync with pyro.   I think if we are going to use "mech config" files (whether they are CTI or YAML, or whatever) that continue to evolve with pyro, then pyro should provide some.   Such an arrangement would not prevent us from using pyro's YAML and customizing it for our own purposes either.

That brings me to another pesky user-feature-request:
Let mirgecom just bypass needing to know how pyro makes the mechanism class at all.  If we (mirgecom) don't need Cantera at runtime, then we don't need to have a "mech config" at all on the mirgecom side. It would be terrific if we could  just ask pyro for a mechanism class by name.  Then it would be none of mirgecom's business whether the class was generated on-the-fly, or pre-generated, or using Cantera or any other package.
   
> > provide pre-generated classes
> 
> I still do not understand the technical reason for pre-generation. But I do understand its substantial technical disadvantages. Care to elaborate?

Pre-generated mechanisms are required to support development of pyro-based transport models, where development versions of Cantera and pyro were used to generate different mech classes.  This was a much-desired feature for our 2 users who had been tinkering with this.    Now that conda actually installs a compatible version (I think?), this motivating factor has diminished a bit.

It is not strictly necessary for us to use pre-generated mechanisms for the prediction (or any other driver).  We chose to do it in the prediction drivers because switching to an all-pyro-based init and pre-generated uiuc mechanism (from _pyro_'s own uiuc.cti file) had a profoundly positive affect on the performance of the prediction flow initialization.  This was part of the effort to reduce the complexity of the initialization routines.

MTCam:
> > provide pre-generated classes
> 
> I still do not understand the technical reason for pre-generation. But I do understand its substantial technical disadvantages. Care to elaborate?

@inducer: Per our discussion in this morning's dev meeting, this (/illinois-ceesd/mirgecom#683) removes all pre-generated anything from mirgecom, and incidentally fixes the case we ran into where *this* PR made a change to mech config format that broke mirgecom.   I am not doing this begrudgingly; i've been won over on this issue.

pyrometheus/chem_expr.py
```diff
@@ -159,15 +159,16 @@ def gen(c, t):
 # {{{ Data-handling helper
 
 def _zeros_like(argument):
-    # FIXME: This mishandles NaNs.
+    # Fixme: This mishandles NaNs.
```
inducer, line 162: The all-caps version is actually intentional and treated as special for example by Vim.
```suggestion
    # FIXME: This mishandles NaNs.
```

test/mechs/sandiego.yaml
```diff
@@ -0,0 +1,225 @@
+generator: cti2yaml
```
inducer, line 1: Given that we now have the `.yaml` versions, do we still need the `.cti` versions?

pyrometheus/chem_expr.py
```diff
@@ -159,15 +159,16 @@ def gen(c, t):
...
     return 0 * argument
 
 # }}}
 
 # {{{ Equilibrium constants
 
 
-def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction, gibbs_rt):
+def equilibrium_constants_expr(sol: ct.Solution, react: ct.Reaction,
+                               reaction_index, gibbs_rt):
```
inducer, line 171: The arguments to this seem redundant now, and redundancy breeds inconsistency. Drop the `react` parameter?

pyrometheus/chem_expr.py
```diff
@@ -241,14 +242,34 @@ def troe_falloff_expr(react: ct.Reaction, t):
     :returns: The Troe falloff center expression for reaction *react* in terms of the
         temperature *t* as a :class:`pymbolic.primitives.Expression`
     """
-    troe_params = react.falloff.parameters
+    if react.uses_legacy:
+        from warnings import warn
+        warn("New Cantera implementation retrieves 'FalloffRate' objects "
+             "using the rate 'property'", DeprecationWarning)
```
inducer, line 248: - What is "new Cantera"? (Be precise: version numbers and such.)
- Good messages are actionable: what should the user do to fix this?
- Deprecations typically have a fuse ("when will this stop working?"). Specify it, if it exists, or specify that it doesn't exist.
- If the message is a full sentence, it needs punctuation.

pyrometheus/chem_expr.py
```diff
@@ -257,7 +278,15 @@ def falloff_function_expr(react: ct.Reaction, i, t, red_pressure, falloff_center
...
-    if react.falloff.falloff_type == "Troe":
+    if react.uses_legacy:
+        falloff_type = react.falloff.falloff_type
+        from warnings import warn
+        warn("New Cantera implementation retrieves 'FalloffRate' objects "
+             "using the rate 'property'", DeprecationWarning)
+    else:
+        falloff_type = react.rate.type
+
+    if falloff_type == "Troe":
```
inducer, line 289: I know that this is old code, but this is not a good pattern. Instead, act only on expected values and raise an error on unexpected ones.

pyrometheus/chem_expr.py
```diff
@@ -241,14 +242,34 @@ def troe_falloff_expr(react: ct.Reaction, t):
...
+        if react.rate.type == "Troe":
+            troe_params = react.rate.falloff_coeffs
+        else:
+            # This isn't a Troe function
+            return 1
+
     troe_1 = (1.0-troe_params[0])*p.Variable("exp")(-t/troe_params[1])
     troe_2 = troe_params[0]*p.Variable("exp")(-t/troe_params[2])
-    if troe_params[3]:
+    if len(troe_params) == 4:
```
inducer, line 268: Act only on expected values and raise an error on unexpected ones.

pyrometheus/chem_expr.py
```diff
@@ -241,14 +242,34 @@ def troe_falloff_expr(react: ct.Reaction, t):
     :returns: The Troe falloff center expression for reaction *react* in terms of the
         temperature *t* as a :class:`pymbolic.primitives.Expression`
     """
-    troe_params = react.falloff.parameters
+    if react.uses_legacy:
+        from warnings import warn
+        warn("New Cantera implementation retrieves 'FalloffRate' objects "
+             "using the rate 'property'", DeprecationWarning)
+        if react.falloff.falloff_type == "Troe":
```
inducer, line 249: Raise an error on unexpected values.

pyrometheus/chem_expr.py
```diff
@@ -241,14 +242,34 @@ def troe_falloff_expr(react: ct.Reaction, t):
...
+                troe_params = react.falloff.parameters
+            else:
+                troe_params = react.falloff.parameters[:-1]
+
+        else:
+            # This isn't a Troe function
+            return 1
+
+    else:
+        if react.rate.type == "Troe":
```
inducer, line 260: Act only on expected values and raise an error on unexpected ones.

pyrometheus/codegen/python.py
```diff
@@ -466,12 +471,18 @@ def gen_thermochem_code(sol: ct.Solution) -> str:
...
+                                       for r in sol.reactions()])),
         non_falloff_reactions=list(compress(sol.reactions(),
                                             [not isinstance(r, ct.FalloffReaction)
                                              for r in sol.reactions()])),
         three_body_reactions=list(compress(sol.reactions(),
                                            [isinstance(r, ct.ThreeBodyReaction)
                                             for r in sol.reactions()])),
+        three_body_indices=list(compress(range(sol.n_reactions),
+                                         [isinstance(r, ct.ThreeBodyReaction)
+                                         for r in sol.reactions()]))
```
inducer, line 485: Rewrite these as list comprehensions of tuples with a conditional:
```
[(i, react) for i, react in enumerate(sol.reactions()) if isinstance(react, ...)]
```
This will avoid the need for separate `indices` arrays.

inducer (commented):
Thanks! LGTM after these are addressed.

test/mechs/sandiego.yaml
MTCam: It might be nice to have one just to test that they are still supported?

inducer: Fair point. I don't see a good reason to keep all of them in both formats though.

pyrometheus/chem_expr.py
ecisneros8: We still need `react` because we still get values out of it, such as the indices of its reactants and products:

[https://github.com/pyrometheus/pyrometheus/blob/437f5663b346283ba81cee51421b5cf274eb1998/pyrometheus/chem_expr.py#L178](`indices_reac = [sol.species_index(sp) for sp in react.reactants]`)

The problem is that, to the best of my knowledge, a Cantera-2.6 `Reaction` does not know the reaction's "global" index in a mechanism (in Cantera 2.5, it had `ID`).

inducer: > The problem is that, to the best of my knowledge, a Cantera-2.6 Reaction does not know the reaction's "global" index in a mechanism (in Cantera 2.5, it had ID).

But it looks like you'll be able to get the reaction given the `sol` and the index? (Or am I missing something?)

pyrometheus/chem_expr.py
inducer: (I don't recall a situation where I found a warning or error message to have too much explanation or context.)

pyrometheus/chem_expr.py
ecisneros8: Indeed, that's sounds like the right way to do it... I hadn't had coffee when I looked at this this morning.

pyrometheus/chem_expr.py
ecisneros8: Sorry to ask for further clarification, but does this mean that I should not act (generate expressions) on legacy (cti) reaction patterns? That would lead to mirgecom test failure, unless mirgecom itself has already been updated too. I think there is something I'm still missing.

inducer: What I mean here is the following pattern:
```py
if falloff_type == "some_known_value":
    do the thing
else:
    # silently assume it's something else
    do the other thing
```
A more defensive way to do the same thing:
```py
if falloff_type == "some_known_value":
    do the thing
elif falloff_type == "another_known_value":
    do the other thing
else:
    raise ValueError(f"unexpected value of 'falloff_type': '{falloff_type}'")
```

ecisneros8:
Implemented changes suggested by @inducer, but downstream mirgecom is failing. I looked at the main script but couldn't find helpful information. I ran test_codegen.py locally with .cti input files, and these pass, so I don't know what could be causing this. @MTCam

MTCam:
> Implemented changes suggested by @inducer, but downstream mirgecom is failing. I looked at the main script but couldn't find helpful information. I ran test_codegen.py locally with .cti input files, and these pass, so I don't know what could be causing this. @MTCam

Ah, no worries here @ecisneros8 .  @inducer is already on top of fixing that up in illinois-ceesd/mirgecom#686

inducer:
Thanks! This LGTM.

In accordance with the discussion above, I got rid of all the `.cti`s except one and added a test to ensure that we can still read CTI-format files.

inducer (approved):
LGTM!

PR #53, opened 2022-06-07
Stop using deprecated `np.object` in generated Python

inducer:


inducer:
There's nothing much to this. Merging without review.

PR #56, opened 2022-07-18
Paper Examples and Concentrations Array

ecisneros8:
This PR changes the generated code inside `get_concentrations` from 
`self.iwts * density * mass_fractions` 
to 
`self._pyro_make_array([self.iwts[0] * density * mass_fractions[0], ..., self.iwts[${num_species}] * density * mass_fractions[${num_species}]])`

This enables evaluating rates on grids (meaning the inputs can be multi-dimensional arrays, not point wise quantities), so a corresponding test is added. It also enables tracing the code with Pytato, but no test was added yet.

### DIFF
--- a/pyrometheus/codegen/python.py
+++ b/pyrometheus/codegen/python.py
@@ -260,7 +260,11 @@ class Thermochemistry:
                 )
 
     def get_concentrations(self, rho, mass_fractions):
-        return self.iwts * rho * mass_fractions
+        return self._pyro_make_array([
+            % for i in range(sol.n_species):
+            self.iwts[${i}] * mass_fractions[${i}] * rho,
+            % endfor
+                ])
 
     def get_mass_average_property(self, mass_fractions, spec_property):
         return sum([mass_fractions[i] * spec_property[i] * self.iwts[i]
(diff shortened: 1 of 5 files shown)


pyrometheus/codegen/python.py
```diff
@@ -260,7 +260,11 @@ def get_mix_molecular_weight(self, mass_fractions):
                 )
 
     def get_concentrations(self, rho, mass_fractions):
-        return self.iwts * rho * mass_fractions
+        return self._pyro_make_array([
+            % for i in range(sol.n_species):
+            self.iwts[${i}] * mass_fractions[${i}] * rho,
+            % endfor
+                ])
```
MTCam, line 267: I wonder if this will obviate mirgecom's customization of this function.  I'll test it out tomorrow.

ecisneros8: Hi @MTCam, following up on this. Did it obviate mirgecom's customization?

MTCam: No, it doesn't.  Not quite sure what I was thinking here.  MC squishes the concentrations below a certain level (typically ~1e-10) to 0.   The big problem is that very small mass fractions (~1e-13ish) result in significant reaction rates, which run the mass fractions significantly negative over a simulation time step.  So now we ignore everything below a certain threshold - which is set up in our customization of `get_concentrations`.

ecisneros8:
These features have already been merged via other branches.

PR #73, opened 2023-08-18
Add functions get_specific_heat_ratio and get_soundspeed

jonathan-macart:
This commit adds:
1. A function to return the specific heat ratio ($\gamma = c_p/c_v$) without evaluating the NASA polynomials twice as a naiive approach would. 
2. A function to return the speed of sound ($a = \sqrt{\gamma R T}$), which is needed in several places within compressible flow solvers.

PR #75, opened 2023-09-07
Updates for Compatibility with the new Cantera 3.0 Release

ecisneros8:
This PR updates the Pyrometheus code template and `chem_expr` module to ensure compatibility with the new Cantera release. 

The principal change in Cantera is that all reactions in a `ct.Solution` are of generic type `ct.Reaction`, no longer of any specific type (e.g., `ct.FalloffReaction`). Reaction types are now identified via the `ct.Reaction.reaction_type` interface. Pyrometheus has been updated accordingly to distinguish between elementary, falloff, and third-body reactions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
                 curl -L -O https://tiker.net/ci-support-v0
                 . ./ci-support-v0
 
-                git clone -b use_cantera30 "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                git clone -b "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
                 cd "$DOWNSTREAM_PROJECT"
                 echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
                 edit_requirements_txt_for_downstream_in_subdir
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
                 curl -L -O https://tiker.net/ci-support-v0
                 . ./ci-support-v0
 
-                git clone "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                git clone -b use_cantera30 "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
                 cd "$DOWNSTREAM_PROJECT"
                 echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
                 edit_requirements_txt_for_downstream_in_subdir
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
                 # "3" is intentional: It's suppposed to be conda's default
                 # Python version number, which seems to have moved past
                 # 3.9 as of 2022-02-20.
-                python_version: ["3.9", "3"]
+                python_version: ["3.9", "3.11"]
         runs-on: ubuntu-latest
         env:
             PYTHON_VERSION: ${{ matrix.python_version }}
--- a/pyrometheus/chem_expr.py
+++ b/pyrometheus/chem_expr.py
@@ -258,7 +258,7 @@ def troe_falloff_expr(react: ct.Reaction, t):
     :returns: The Troe falloff center expression for reaction *react* in terms of the
         temperature *t* as a :class:`pymbolic.primitives.Expression`
     """
-    if hasattr("uses_legacy", react) or not react.uses_legacy:
+    if "uses_legacy" not in dir(react) or not react.uses_legacy:
         if isinstance(react.rate, ct.TroeRate):
             troe_params = react.rate.falloff_coeffs
         elif isinstance(react.rate, ct.LindemannRate):
@@ -303,7 +303,6 @@ def falloff_function_expr(react: ct.Reaction, i, t, red_pressure, falloff_center
         of the temperature *t*, reduced pressure *red_pressure*, and falloff center
         *falloff_center* as a :class:`pymbolic.primitives.Expression`
     """
-    print(react.uses_legacy)
     if "uses_legacy" not in dir(react) or not react.uses_legacy:
         # Assume Cantera 3.0
         falloff_type = react.reaction_type.split("-")[1]
--- a/pyrometheus/chem_expr.py
+++ b/pyrometheus/chem_expr.py
@@ -227,7 +227,7 @@ def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
         of the species concentrations *c* as a
         :class:`pymbolic.primitives.Expression`
     """
- 
+
     efficiencies = [react.third_body.efficiencies[sp]
                     for sp in react.third_body.efficiencies]
     indices_nondef = [sol.species_index(sp) for sp
--- a/pyrometheus/chem_expr.py
+++ b/pyrometheus/chem_expr.py
@@ -229,7 +229,8 @@ def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
         of the species concentrations *c* as a
         :class:`pymbolic.primitives.Expression`
     """
-    if isinstance(react, ct.ThreeBodyReaction):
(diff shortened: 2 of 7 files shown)


pyrometheus/chem_expr.py
```diff
@@ -229,9 +229,23 @@ def third_body_efficiencies_expr(sol: ct.Solution, react: ct.Reaction, c):
...
         :class:`pymbolic.primitives.Expression`
     """
-    efficiencies = [react.efficiencies[sp] for sp in react.efficiencies]
-    indices_nondef = [sol.species_index(sp) for sp in react.efficiencies]
-    indices_default = [i for i in range(sol.n_species) if i not in indices_nondef]
+    if isinstance(react, ct.ThreeBodyReaction):
+        from warnings import warn
+        warn("The 'Reaction.efficiencies' interface is deprecated and "
+             "will be removed after Cantera 3. Access efficiencies via "
+             "'react.third_body.efficiencies' instead")
```
inducer, line 236: I'm not sure we need to warn here. There's no thing actionable for the pyro user here.

pyrometheus/chem_expr.py
```diff
@@ -243,10 +257,10 @@ def troe_falloff_expr(react: ct.Reaction, t):
     :returns: The Troe falloff center expression for reaction *react* in terms of the
         temperature *t* as a :class:`pymbolic.primitives.Expression`
     """
-    if not react.uses_legacy:
-        if react.rate.type == "Troe":
+    if "uses_legacy" not in dir(react) or not react.uses_legacy:
```
inducer, line 260:
```suggestion
    if hasattr(react, "uses_legacy") or not react.uses_legacy:
```
(similar elsewhere)

pyrometheus/codegen/python.py
```diff
@@ -476,6 +476,24 @@ def gen_thermochem_code(sol: ct.Solution) -> str:
     adhering to the :class:`~pyrometheus.thermochem_example.Thermochemistry`
     interface.
     """
+    if not all((isinstance(r, ct.Reaction) for r in sol.reactions())):
+        # Cantera version < 3.0
+        from warnings import warn
+        warn("Specific reaction types (e.g., ct.FalloffReaction) are "
+             "deprecated in Cantera 3, where all in a 'ct.Solution' "
+             "object are of generic 'ct.Reaction' type. "
+             "Identify reactions using 'ct.Reaction.reaction_type' insteady")
```
inducer, line 485: As above.

pyrometheus/codegen/python.py
```diff
@@ -427,7 +427,7 @@ def get_fwd_rate_coefficients(self, temperature, concentrations):
         ones = self._pyro_zeros_like(temperature) + 1.0
         k_fwd = [
         %for react in sol.reactions():
-        %if isinstance(react, ct.FalloffReaction):
+        %if react.equation in [r.equation for _, r in falloff_reactions]:
```
inducer, line 430: This idiom is kind of inefficient. It seems like we could just test `react` instead?

test/test_codegen.py
```diff
@@ -132,7 +136,22 @@ def test_get_rate_coefficients(mechname, usr_np):
...
+        # It seems like Cantera 3.0 has bumped the third-body efficiency
+        # factor from the rate coefficient to the rate of progress
+        for i, r in three_body_reactions:
+            eff = np.sum([c[sol.species_index(sp)]
+                          * r.third_body.efficiencies[sp]
+                          for sp in r.third_body.efficiencies])
+            eff += np.sum([c[sp] for sp in range(sol.n_species)
+                           if sol.species_name(sp) not in
+                           r.third_body.efficiencies])
+            k_ct[i] *= eff
```
inducer, line 149: Would this maintain compatibility with Cantera < 3?

inducer (commented):
Thanks for working on this! My main question is whether we're hoping to maintain compatbility with cantera < 3. Some of the code is written as if that's the case, some other bits I'm not sure. If yes (i.e. maintain compatbility): why? The simpler thing to do might be to just make the jump (and update the dependency version in `setup.py`.

inducer:
Unsubscribing... @-mention or request review once it's ready for a look or needs attention.

pyrometheus/codegen/python.py
ecisneros8: That was my initial approach but found that, for example, for the San Diego yaml in `mechs/test`:

`sol.reaction(9).equation == falloff_reactions[0].equation`

is true, so the 10th reaction is Falloff, and it is the first in the `falloff_reactions` list, but

`sol.Reaction(9) == falloff_reactions[0]`

is false, so testing `react in falloff_reactions` does not help here. I haven't been able to pin down what specific property makes them appear different upon comparison.

test/test_codegen.py
ecisneros8: Thanks for noticing this. It will not. I will fix accordingly.

pyrometheus/chem_expr.py
ecisneros8: But we do have some users that modify the codegen. Indeed, @jonathan-macart has one such PR #73 !

ecisneros8:
Thanks for reviewing this PR! That is a good point, and a simple way forward. 

I think that the strongest case for maintaining compatibility is to support our users that still use Cantera 2.6 in their toolchains. After all, Cantera 3 was release just 3 weeks ago or so. If we choose to support both versions, we will have to plan on when to drop < 3, correct?

tulioricci:
Hi @ecisneros8, any issues towards merging this PR? Anything I could help with?

ecisneros8:
Hi @tulioricci, thanks for pointing us back at this. We didn't merge this right away because the downstream Mirgecom test failed. My recollection is that it had to do with [building mpi4py](https://github.com/pyrometheus/pyrometheus/actions/runs/6426874713/job/17451685979#step:3:1303) during environment installing for the test. 

On the Pyro side everything is ready (but please confirm @inducer). If you can make the downstream test pass, it's a go. As a hint that may help, the Python version for tests is restricted to 3.11, because that's what Cantera supported when I created the PR.

tulioricci:
Mirge-com tests will not pass while there is no mixture-averaged transport model inside pyrometheus.

tulioricci:
> Hi @tulioricci, thanks for pointing us back at this. We didn't merge this right away because the downstream Mirgecom test failed. My recollection is that it had to do with [building mpi4py](https://github.com/pyrometheus/pyrometheus/actions/runs/6426874713/job/17451685979#step:3:1303) during environment installing for the test.

Mirge-com tests will not pass while there is no mixture-averaged transport model inside pyrometheus.

matthiasdiener:
> Hi @tulioricci, thanks for pointing us back at this. We didn't merge this right away because the downstream Mirgecom test failed. My recollection is that it had to do with [building mpi4py](https://github.com/pyrometheus/pyrometheus/actions/runs/6426874713/job/17451685979#step:3:1303) during environment installing for the test.

#77 should fix the mpi4py-related error.

test/test_codegen.py
```diff
@@ -110,7 +110,8 @@ def test_generate_mechfile(lang_module, mechname):
         print(code, file=mech_file)
 
 
-@pytest.mark.parametrize("mechname", ["uiuc.yaml", "sandiego.yaml", "uiuc.cti"])
+@pytest.mark.parametrize("mechname", ["uiuc.yaml", "sandiego.yaml",
+                                      "uiuc.yaml"])
```
tulioricci, line 114: Duplicated

test/test_codegen.py
tulioricci, line 149: I have the impression that mirge-com hates *= and +=. Please, write this (and everywhere else) as `k_ct[i] = k_ct[i]*eff`

.github/workflows/ci.yml
```diff
@@ -104,7 +104,7 @@ jobs:
                 curl -L -O https://tiker.net/ci-support-v0
                 . ./ci-support-v0
 
-                git clone "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                git clone -b use_cantera30 "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
```
tulioricci, line 107: Let's not forget to undo this

tulioricci (approved):
LGTM. Just address my 2 comments and merge at your will.

I tested this in mirge-com as well and it seems to be working perfectly.

PR #79, opened 2024-01-06
Add pressure terms to entropy

tulioricci:


pyrometheus/chem_expr.py
```diff
@@ -301,12 +300,13 @@ def rate_coefficient_expr(rate_coeff: ct.Arrhenius, t):
...
     b = rate_coeff.temperature_exponent
     t_a = rate_coeff.activation_energy/ct.gas_constant
-    if t_a == 0:
-        # Weakly temperature-dependent rate
-        return a * t**b
-    else:
-        # Modified Arrhenius
-        return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
+    return p.Variable("exp")(np.log(a)+b*p.Variable("log")(t)-t_a/t)
+#    if t_a == 0:
```
ecisneros8, line 304: Why is this commented out?

pyrometheus/chem_expr.py
```diff
@@ -280,13 +280,12 @@ def equilibrium_constants_expr(sol: ct.Solution, reaction_index, gibbs_rt):
             for indices_prod_i, nu_prod_i in zip(indices_prod, nu_prod))
 
     # Check if reaction is termolecular
-    sum_nu_net = sum(nu_prod) - sum(nu_reac)
+    sum_nu_net = sum(nu_reac) - sum(nu_prod)
```
ecisneros8, line 283: Any reason for sign changes here and line 285?

pyrometheus/chem_expr.py
```diff
@@ -280,13 +280,12 @@ def equilibrium_constants_expr(sol: ct.Solution, reaction_index, gibbs_rt):
...
     if sum_nu_net != 0:
-        return sum_p - sum_r - sum_nu_net*p.Variable("c0")
+        return sum_p - sum_r + sum_nu_net*p.Variable("c0")
```
ecisneros8, line 285: Related to the previous comment...

pyrometheus/codegen/python.py
```diff
@@ -321,6 +327,33 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
 
+    def get_mixture_entropy_mass(self, pressure, temperature, mass_fractions):
+        mmw = self.get_mix_molecular_weight(mass_fractions)
+        return 1.0/mmw * self.get_mixture_entropy_mole(pressure, temperature, <%
+                                                       %>mass_fractions)
+
+    def get_mole_average_property(self, mass_fractions, spec_property):
+        mmw = self.get_mix_molecular_weight(mass_fractions)
+        mole_fracs = self.get_mole_fractions(mmw, mass_fractions)
+        return sum([mole_fracs[i] * spec_property[i]
```
ecisneros8, line 338: Should this not be a `usr_np.sum`?

pyrometheus/codegen/python.py
```diff
@@ -133,9 +134,14 @@ class Thermochemistry:
     .. automethod:: get_mix_molecular_weight
     .. automethod:: get_concentrations
     .. automethod:: get_mole_fractions
+    .. automethod:: get_mass_average_property
+    .. automethod:: get_mole_average_property
```
ecisneros8, line 138: Why are mole-averaged properties needed?

pyrometheus/codegen/python.py
```diff
@@ -307,9 +313,9 @@ def get_mixture_specific_heat_cp_mass(self, temperature, mass_fractions):
         return self.gas_constant * cpmix
 
     def get_mixture_specific_heat_cv_mass(self, temperature, mass_fractions):
-        cp0_r = self.get_species_specific_heats_r(temperature) - 1.0
-        cpmix = self.get_mass_average_property(mass_fractions, cp0_r)
-        return self.gas_constant * cpmix
+        cv0_r = self.get_species_specific_heats_r(temperature) - 1.0
+        cvmix = self.get_mass_average_property(mass_fractions, cv0_r)
```
ecisneros8, line 317: Should we use `cv_mix` instead of `cvmix`? I prefer the former over the latter, but I can't recall how the notation throughout the rest of the template.

pyrometheus/codegen/python.py
```diff
@@ -321,6 +327,33 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+
+    def get_mixture_enthalpy_mole(self, temperature, mass_fractions):
+        h0_rt = self.get_species_enthalpies_rt(temperature)
+        hmix = self.get_mole_average_property(mass_fractions, h0_rt)
+        return self.gas_constant * temperature * hmix
+
+    def get_mixture_entropy_mole(self, pressure, temperature, mass_fractions):
+        mmw = self.get_mix_molecular_weight(mass_fractions)
+        # necessary to avoid nans in the log function below
+        x = self.usr_np.where(
```
ecisneros8, line 349: I am uneasy about this type of stuff being part of the template. Generally, these numerical tricks can be worked in by overriding individual functions of the generated code.  Particularly, the resulting mole fractions `x` do not correspond to the mass fractions passed as inputs.

pyrometheus/codegen/python.py
```diff
@@ -321,6 +327,33 @@ def get_mixture_internal_energy_mass(self, temperature, mass_fractions):
...
+            self.usr_np.less(self.get_mole_fractions(mmw, mass_fractions), 1e-16),
+            1e-16, self.get_mole_fractions(mmw, mass_fractions))
+        s0_r = self.get_species_entropies_r(pressure, temperature)
+        smix = self.get_mole_average_property(mass_fractions, s0_r)
+        xmix = self.get_mole_average_property(mass_fractions, self.usr_np.log(x))
+        return self.gas_constant * (smix - xmix)
```
ecisneros8, line 355: I don't like this notation. `xmix` does not right away indicate an entropy. We should have something like `s_t_mix` and `s_x_mix` to denote the individual contributions to the entropy.

pyrometheus/codegen/python.py
```diff
@@ -335,23 +368,24 @@ def get_species_enthalpies_rt(self, temperature):
             % endfor
             ])
 
-    def get_species_entropies_r(self, temperature):
+    def get_species_entropies_r(self, pressure, temperature):
         return self._pyro_make_array([
             % for sp in sol.species():
-            ${cgm(ce.poly_to_entropy_expr(sp.thermo, "temperature"))},
+            ${cgm(ce.poly_to_entropy_expr(sp.thermo, "temperature"))}
+            - self.usr_np.log(pressure/101325.0),
```
ecisneros8, line 375: `101325.0` -> `self.one_atm`

pyrometheus/codegen/python.py
```diff
@@ -335,23 +368,24 @@ def get_species_enthalpies_rt(self, temperature):
...
         h0_rt = self.get_species_enthalpies_rt(temperature)
-        s0_r = self.get_species_entropies_r(temperature)
+        s0_r = self.get_species_entropies_r(pressure, temperature)
         return h0_rt - s0_r
 
-    def get_equilibrium_constants(self, temperature):
+    def get_equilibrium_constants(self, pressure, temperature):
         rt = self.gas_constant * temperature
-        c0 = self.usr_np.log(self.one_atm / rt)
+        c0 = self.usr_np.log(pressure / rt)
```
ecisneros8, line 386: Does the `self.usr_np.log(pressure / rt)` along with the `np.log(x)` term in the mixture entropy not reduce to `self.usr_np.log(self.one_atm / rt)`? Somehow I think it is related to why we represented it in this way to begin with. I looked at the math of the `\sum_i x_i log(x_i)` term a long time ago for a couple of different reasons, though I can't remember for sure if it impacts this term.

pyrometheus/codegen/python.py
```diff
@@ -335,23 +368,24 @@ def get_species_enthalpies_rt(self, temperature):
...
 
-        g0_rt = self.get_species_gibbs_rt(temperature)
+        g0_rt = self.get_species_gibbs_rt(pressure, temperature)
```
ecisneros8, line 388: Here, and in the lines that follow, let's make sure we have consistency in the way thermo variables are ordered as function arguments. (It may be correct already.)

test/test_codegen.py
```diff
@@ -113,53 +111,6 @@ def test_generate_mechfile(lang_module, mechname):
...
-
-        print(k_ct)
-        print()
-        print(k_pm)
-        print()
-        print(np.abs((k_ct-k_pm)/k_ct))
-        assert np.linalg.norm((k_ct-k_pm)/k_ct, np.inf) < 1e-14
-    return
-
-
```
ecisneros8, line 162: Why is this test being deleted?

test/test_codegen.py
```diff
@@ -199,64 +150,108 @@ def test_get_pressure(mechname, usr_np):
...
-        for i, reaction in enumerate(sol.reactions()):
-            if reaction.reversible:
-                keq_err = np.abs((keq_pm[i] - keq_ct[i]) / keq_ct[i])
-                print(f"keq_err = {keq_err}")
-                assert keq_err < 1.0e-13
-        # keq_pm_test = keq_pm[1:]
-        # keq_ct_test = keq_ct[1:]
-        # keq_err = np.linalg.norm((keq_pm_test - keq_ct_test) / keq_ct_test, np.inf)
-        # assert keq_err < 1.0e-13
+    oxidizer = {"O2": 1.0, "N2": 3.76}
```
ecisneros8, line 178: Why are we excluding this from the test?

pyrometheus/codegen/python.py
```diff
@@ -423,48 +456,62 @@ def get_falloff_rates(self, temperature, concentrations, k_fwd):
...
+        k_falloff = self.get_falloff_rates(temperature, concentrations)
+        %endif
         k_fwd = [
-        %for react in sol.reactions():
-        %if react.equation in [r.equation for _, r in falloff_reactions]:
-            0*temperature,
+        %for i, react in enumerate(sol.reactions()):
+        %if react.reaction_type.startswith("falloff"):
+        %for k, (j, react) in enumerate(falloff_reactions):
+        %if i == j:
```
ecisneros8, line 471: Is there a way of doing this without nested loops?

ecisneros8 (commented):
@tulioricci thanks for working on this. I left some comments that I think are easy to address. 

One of my comments concerns the `log(p_atm / rt)` and `sum_i x_i log(x_i)` terms in the equilibrium constants and molar mixture entropy. Hopefully it is clear enough. If not, let me know and we can meet.

pyrometheus/chem_expr.py
tulioricci: I commented out the previous implementation, left it there as a reminder. I personally prefer having the same expression for both cases. But no strong opinion on this.

pyrometheus/chem_expr.py
tulioricci: Obsessive compulsive disorder. I dislike "- -1.0" expressions.

pyrometheus/codegen/python.py
tulioricci: In this case maybe. I say that because some np.sum were yielding wrong results in mirge-com: it would sum over all elements of all arrays and return a single number instead of an array. *This is something we have to check in pyrometheus with grids!*

pyrometheus/codegen/python.py
tulioricci: I am just replacing cp -> cv in the "cv function". If you wanna change to cv_mix, we have to change cp_mix in the previous function too.

pyrometheus/codegen/python.py
tulioricci: The issue is that the code will simply crash. There is no log of negative species and this is not something we can put outside of pyrometheus.

pyrometheus/codegen/python.py
tulioricci: :+1:

pyrometheus/codegen/python.py
tulioricci: :+1:

pyrometheus/codegen/python.py
tulioricci: I don't get what you are meaning, exactly. Could you elaborate, please?

pyrometheus/codegen/python.py
tulioricci: :+1:

test/test_codegen.py
tulioricci: We are testing the reactions in a second test down below. I don't see why this one is necessary.

pyrometheus/codegen/python.py
tulioricci: JAX complains of indexations. So modifying only the reactions that are falloff or three-body individually AFTER the pyro_array expression makes JAX fail.

I won;t deny that this is ugly, we can potentially do it nicer, but currently it works.

pyrometheus/codegen/python.py
tulioricci: Tbh, I dont recall exactly. I thought it was important when I started but doesn't look like it anymore. I will double check. In case we remove this will also eliminate some comments you made below.

I will get back on this soon and give an update.

test/test_codegen.py
tulioricci: Just making it from a temperature loop to a reactor case. You will notice that the same properties are being tested below. Lemme know if I skipped any.

pyrometheus/codegen/python.py
ecisneros8: Agreed on Jax, I actually prefer in general what you have done here to avoid `k_fwd[${I}]` (or whatever it was) in `get_falloff_rates`. However, my concern is the template syntax, specifically the nested `%for i, ...` and `%for k, ...`, along with the `%if i == j`. Let me think how to simplify.

pyrometheus/codegen/python.py
ecisneros8: Ok, thanks for clarifying. We should modify `get_density` for consistency then.

pyrometheus/codegen/python.py
ecisneros8: `self.usr_np.sum(..., axis=0)`?

pyrometheus/chem_expr.py
ecisneros8: Understood, but `nu_net` often appears in references as `nu_prod - `nu_reac`, so let's keep it as such.

pyrometheus/chem_expr.py
ecisneros8: I think that if we always return the full expression, we will always end up evaluating an `exp` with a call to a `log` inside, at least in Python, making the evaluation more expensive than it needs to be. Maybe these get removed when translating to OpenCL using Pytato, but we have users that work fully in Python.

pyrometheus/codegen/python.py
ecisneros8: Ok, happy to keep them. I just wasn't sure if mole-based properties are needed anywhere in the formulation.

pyrometheus/codegen/python.py
ecisneros8: Great, thanks!

pyrometheus/codegen/python.py
ecisneros8: I'm confused now. Do we keep mole-based properties or not? Or are they now in `test_codegen.py`? Let me take a look, but I don't think they belong there. We can have an auxiliary file that adds them to the generated class (I can do that).

test/test_codegen.py
ecisneros8: But are we testing the thermo properties of individual species in the reactor case? The deleted portion tests the individual polynomials, which are (except for s maybe) only functions of T.

ecisneros8:
Hi @tulioricci, can you remind me what the status of this PR is? Thanks in advance!

PR #81, opened 2024-04-19
Correct Limit for Falloff Function

ecisneros8:
This PR implements the correct behavior of falloff functions in the limiting case of zero reduced pressure. This behavior can arise for Falloff reactions with only species as a third-body partner e.g., 

`H + O2 (+ Ar) <=> HO2 (+ Ar)`,

when the third-body (Ar) concentration is zero. 

To achieve the right behavior, this PR separates the expression for the falloff factor---where the problematic `log10(reduced_pressure)` appears---from the expression for the falloff function. The code template is modified accordingly. To test the generated code, the Hong mechanism, which contains the aforementioned reaction, is added to the verification suite.

### DIFF
--- a/pyrometheus/chem_expr.py
+++ b/pyrometheus/chem_expr.py
@@ -29,7 +29,8 @@ Internal Functionality
 .. autofunction:: equilibrium_constants_expr
 .. autofunction:: rate_coefficient_expr
 .. autofunction:: third_body_efficiencies_expr
-.. autofunction:: troe_falloff_expr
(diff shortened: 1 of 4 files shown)


inducer:
This looks good I think, but we should clean up the various aspects of CI failure before merging this. See #82.

inducer:
Rebased this.

inducer (approved):

ecisneros8:
Thanks @inducer !

PR #82, opened 2024-04-19
Drop mirgecom downstream CI

inducer:
cc @ecisneros8 @MTCam 

As far as I can see, there is no intention to keep mirgecom working with pyrometheus main. So I don't see a reason to keep this test around.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,43 +88,4 @@ jobs:
                 python .ci/generate-example-mech.py test/mechs/sandiego.yaml pyrometheus/thermochem_example.py
                 build_docs
 
-    downstream_tests:
-        strategy:
-            matrix:
-                downstream_project: [mirgecom]
-        name: Tests for downstream project ${{ matrix.downstream_project }}
-        runs-on: ubuntu-latest
-        steps:
-        -   uses: actions/checkout@v3
-        -   name: "Main Script"
-            env:
-                DOWNSTREAM_PROJECT: ${{ matrix.downstream_project }}
-            run: |
-                set -x
-                curl -L -O https://tiker.net/ci-support-v0
-                . ./ci-support-v0
-
-                git clone -b "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
-                cd "$DOWNSTREAM_PROJECT"
-                echo "*** $DOWNSTREAM_PROJECT version: $(git rev-parse --short HEAD)"
-                edit_requirements_txt_for_downstream_in_subdir
-                # Avoid slow or complicated tests in downstream projects
-                export PYTEST_ADDOPTS="-k 'not (slowtest or octave or mpi)'"
-                if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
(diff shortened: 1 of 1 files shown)


inducer:
(I'll wait for at least one of you to chime in before merging this.)

MTCam:
> (I'll wait for at least one of you to chime in before merging this.)

As near as I can tell, [the actual CI failure ](https://github.com/pyrometheus/pyrometheus/actions/runs/8678203996/job/23794813717#step:3:56) doesn't involve mirgecom at all.   I agree that the upstream tests here are largely irrelevant, but in this failure case it doesn't even look to me like it is getting to mirgecom before it fails.  

Assuming the CI scripting problem can be fixed, could the upstream test just do `pytest mirgecom/test/test_eos.py`?   This would hit the relevant tests only and skip all the other ones (which currently take hour+ to complete).

inducer:
Tried reviving it in #83. I agree the current failure is unrelated, that's due to a `-b` flag that was added in https://github.com/pyrometheus/pyrometheus/commit/152cd69ec7c3bc4b90ea1f092a9c31942eb03a3c for no apparent reason.

inducer:
#83 bombed, as you can see. Which brings us back here.

MTCam:
> #83 bombed, as you can see. Which brings us back here.

Is merging #79 another option that could get the test passing without dropping it?

ecisneros8:
Hi @MTCam and @inducer ,

I do not have a preference on whether to keep the mirgecom test, as I am out on the loop on that code.  The only thing is that...

> Is merging #79 another option that could get the test passing without dropping it?

I am not fully convinced that the changes in #79 are needed. I've been able to simulate related variable-pressure systems without those terms and reproduce published relevant equilibrium temperature and ignition time (I can expand on this if it isn't clear) to within acceptable accuracy (within 1 K for T_eq and 4% ignition time). Yet, @tulioricci had mentioned facing problems without these terms, so certainly further discussion is needed.

If there is no rush to decide on which direction to move, I can look into Tulio's changes and determine whether they are truly warranted over the weekend and early next week---say by Tuesday.

Let me know if this looks reasonable to you, or you need to a decision sooner.

MTCam:
> I am not fully convinced that the changes in #79 are needed. I've been able to simulate related variable-pressure systems without those terms and reproduce published relevant equilibrium temperature and ignition time (I can expand on this if it isn't clear) to within acceptable accuracy (within 1 K for T_eq and 4% ignition time). Yet, @tulioricci had mentioned facing problems without these terms, so certainly further discussion is needed.

In that case, if you are not committed to keeping the test of mirgecom, then I would say drop it.  It is not the responsibility of pyrometheus to keep up with mirgecom. It is the other way around.  The mirgecom test was just a convenience for both of us as an early warning, and it has definitely stopped being that and is now more of an impediment to your development.

While I think the existence of #79 indicates some degree of intent to keep mirgecom working with pyro@main, I also agree that the test should go away (i.e. unless #79 is ready-to-go).

inducer:
OK, then let's get rid of the mirgecom test for now. It's easy enough to bring back.

PR #85, opened 2024-05-03
sync with  main

anderson2981:
Add lower and upper regions for temperature fits with constant Cp. We compute coefficients using the min and max temperature values specified in the NASA7 input coefficients.


ecisneros8:
Somehow this slipped through my radar. This looks useful enough to review. I will come to it as soon as I can.

PR #86, opened 2024-09-21
Fix 64-bit Jax logic

ecisneros8:
This PR fixes a bug that arises from an update to Jax: to enable 64-bit mode, the argument must be boolean, not an integer (as we had before).

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -8,3 +8,4 @@ dependencies:
 - cantera
 - pip
 - jax
+- jaxlib=0.4.31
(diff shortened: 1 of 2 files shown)


ecisneros8:
@inducer As you may have been away, Pyro tests were failing. The failure came from enabling Jax 64-bit mode, which requires a `bool` input after version 0.4.27. The fix is obviously simple. 

However, I could not reconcile testing with Jax 0.4.31 (latest) and Python 3.9, so I pinned the Jax version at 0.4.28. The alternative is to no longer test for Python 3.9 and keep Jax at 0.4.31, but I don't think this would be a good idea until Python 3.9 is deprecated. 

If you think this versioning stuff is sensible, then I'll just go and merge to main.

ecisneros8:
Merging to avoid future failures.

PR #89, opened 2024-10-18
readme fixups

sbryngelson:


ecisneros8 (approved):
Thanks for catching these @sbryngelson !

PR #60, opened 2022-10-07
Transport Properties

ecisneros8:
This PR implements transport properties. It follows from #49 after I accidentally killed it.

PR #92, opened 2024-11-06
Transport Properties (among other things)

henryleberre:
- Reapplies the work done in #60 to be up to date with main.
- Implements transport properties for C++ & Fortran.
- Makes Pyro more modular by creating and exposing a common interface to generate code with. We now have one class implementing the `CodeGenerator` interface for each language we support. These are also available with just an `import pyrometheus as pyro` and don't require digging into the nested `.codegen.<language>` modules. You can still call `pyro.<method>` directly (instead of passing through the `PythonCodeGenerator` class).
- Adds a CLI tool: `pyrometheus`.
```console
usage: pyrometheus [-h] [--version] -l {python,cpp,fortran} -m MECH -o OUTPUT
                   -n NAME [-p PHASE]

Code generation for combustion thermochemistrybased on Cantera.

options:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -l, --lang, --language {python,cpp,fortran}
                        Language to generate code for.
  -m, --mech, --mechanism MECH
                        Path to the mechanism file.
  -o, --output OUTPUT   Path to the output file.
  -n, --name, --namespace NAME
                        Namespace to use for the generated code.
  -p, --phase PHASE     Phase name to use for the generated code.
```
- Adds an example of calling `get_species_binary_mass_diffusivities` with an array of temperatures before extracting the graph of computations for each value produced.


sbryngelson:
@ecisneros8 have you had a chance to look at this

ecisneros8 (approved):
PR approved. Thanks @henryleberre and great work!

PR #97, opened 2025-05-10
Frontier Build Error

DimAdam-01:
When I tried to compile MFC in Frontier I got this error:

"tn-7066 ftn: ERROR S_COMPUTE_PRESSURE, File = ../../../lustre/orion/cfd154/scratch/sbryngelson/runner/actions-runner/_work/MFC/MFC/src/common/m_variables_conversion.fpp, Line = 188 Unsupported accelerator code error: Global in accelerator routine without declare -- inv_molecular_weights"

I renamed the variable inv_weights and revised the implementation of the get_mass_averaged_property subroutine. After those changes, the build completed successfully and the chemistry benchmark tests passed.

### DIFF
--- a/pyrometheus/codegen/fortran.py
+++ b/pyrometheus/codegen/fortran.py
@@ -259,10 +259,10 @@ module ${module_name}
     ${real_type}, parameter :: gas_constant = ${float_to_fortran(ct.gas_constant)}
     ${real_type}, parameter :: molecular_weights(${sol.n_species}) = &
         (/ ${str_np(sol.molecular_weights)} /)
-    ${real_type}, parameter :: inv_molecular_weights(${sol.n_species}) = &
+    ${real_type}, parameter :: inv_weights(${sol.n_species}) = &
         (/ ${str_np(1/sol.molecular_weights)} /)
 
-    !$acc declare create(molecular_weights, inv_molecular_weights)
+    !$acc declare create(molecular_weights, inv_weights)
 
     character(len=12), parameter :: species_names(${sol.n_species}) = &
         (/ ${", ".join('"'+'{0: <12}'.format(s)+'"' for s in sol.species_names)} /)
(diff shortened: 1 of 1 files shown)


pyrometheus/codegen/fortran.py
```diff
@@ -259,10 +259,10 @@ def map_if(self, expr, enclosing_prec):
     ${real_type}, parameter :: gas_constant = ${float_to_fortran(ct.gas_constant)}
     ${real_type}, parameter :: molecular_weights(${sol.n_species}) = &
         (/ ${str_np(sol.molecular_weights)} /)
-    ${real_type}, parameter :: inv_molecular_weights(${sol.n_species}) = &
+    ${real_type}, parameter :: inv_weights(${sol.n_species}) = &
```
ecisneros8, line 262: This change makes the Fortran template inconsistent (wrt to the other templates) in the naming of this variable. Please investigate further why the compiler did not like `inv_molecular_weights`. As discussed in person, it is likely that the problem is no the array name, but how it was used inside a `sum` (line 413).

pyrometheus/codegen/fortran.py
```diff
@@ -410,7 +410,11 @@ def map_if(self, expr, enclosing_prec):
         ${real_type}, intent(in), dimension(${sol.n_species}) :: spec_property
         ${real_type}, intent(out) :: mix_property
 
-        mix_property = sum(inv_molecular_weights*mass_fractions*spec_property)
+        mix_property =  ( &
```
ecisneros8, line 413: This change makes sense. Good catch.

ecisneros8 (changes requested):
Hi @DimAdam-01, thanks for working on this. When possible, variable renaming (in this case `inv_molecular_weights` -> `inv_weights`) must be consistent and well justified across the different languages we support. To avoid creating more work for yourself, please dig deeper into why the name `inv_molecular_weights` gave rise to compilation errors.

sbryngelson:
is this appropriate to be merged now @ecisneros8 ?

ecisneros8:
> is this appropriate to be merged now @ecisneros8 ?

Looks good to me.

ecisneros8 (approved):
@DimAdam-01 thanks for working on this.

PR #116, opened 2026-07-16
Add species creation and destruction rates

sbryngelson:
## Summary

Adds `get_creation_rates` and `get_destruction_rates` to the Python, Fortran, and
C++ code generators, mirroring Cantera's `creation_rates` / `destruction_rates`.
They split the net production rate ω_i into its two non-negative parts:

- **creation_i** — species *i* produced as a product by forward reactions and as a
  reactant by reverse reactions
- **destruction_i** — species *i* consumed as a reactant by forward reactions and as
  a product by reverse reactions

with `creation - destruction == net` by construction. Two small helpers back them,
`get_fwd_rates_of_progress` and `get_rev_rates_of_progress` (the forward/reverse
halves of the existing net rate of progress).

## Motivation

Quasi-steady-state and operator-split stiff integrators (e.g. α-QSS / CHEMEQ2, and
production–loss preconditioners) need the production and loss rates **separately**,
not just the net — the per-species pseudo-first-order loss `L_i = destruction_i / Y_i`
is exactly what those schemes key off. Pyrometheus currently only exposes the net
rate, so downstream codes have to re-derive the split by hand. Concretely, we're
building a GPU-friendly α-QSS reactor integrator on top of Pyrometheus and this is
the one missing primitive.

## Implementation

`chem_expr.py`: `fwd_rate_of_progress_expr` / `rev_rate_of_progress_expr` (the ≥0
halves of `rate_of_progress_expr`) and `creation_rate_expr` / `destruction_rate_expr`
(the same reactant/product stoichiometry walk `production_rate_expr` already uses).
The Python, Fortran, and C++ backends emit the matching routines. All additive — no
change to existing generated output.

## Validation

Generated rates match Cantera to machine precision across mechanisms:

| mechanism | creation err | destruction err | net == c − d |
|---|---|---|---|
| h2o2 (10 sp)  | 2e-17 | 2e-17 | 4e-19 |
| gri30 (53 sp) | 3e-18 | 3e-18 | 1e-19 |

`test_kinetics` is extended to assert both against Cantera's `creation_rates` /
`destruction_rates`, plus the `net == creation - destruction` identity, inside the
existing reactor loop.

### DIFF
--- a/pyrometheus/chem_expr.py
+++ b/pyrometheus/chem_expr.py
@@ -522,21 +522,32 @@ def rev_rate_of_progress_expr(sol: ct.Solution, reaction_index, c,
         * p.Variable("exp")(log_k_eq[reaction_index]) * r_rev
 
 
-def creation_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
+def _species_reaction_stoich(sol: ct.Solution, species):
     """
-    :returns: Species creation (production) rate for *species*: created as a
-        product by forward reactions and as a reactant by reverse reactions.
-        Mirrors Cantera's ``creation_rates``.
+    :returns: ``(idx_reactant, idx_product, nu_reactant, nu_product)`` -- the
+        reaction indices in which *species* appears as a reactant and as a
+        product, with the matching stoichiometric coefficients. Shared by the
+        creation/destruction splits, which differ only in how these are summed.
     """
-    ones = _zeros_like(r_fwd[0]) + 1.0
+    si = sol.species_index(species)
     idx_reactant = [i for i, react in enumerate(sol.reactions())
                     if species in react.reactants]
     idx_product = [i for i, react in enumerate(sol.reactions())
                    if species in react.products]
-    nu_reactant = [sol.reactant_stoich_coeff(sol.species_index(species), i)
-                   for i in idx_reactant]
-    nu_product = [sol.product_stoich_coeff(sol.species_index(species), i)
-                  for i in idx_product]
+    nu_reactant = [sol.reactant_stoich_coeff(si, i) for i in idx_reactant]
+    nu_product = [sol.product_stoich_coeff(si, i) for i in idx_product]
+    return idx_reactant, idx_product, nu_reactant, nu_product
+
+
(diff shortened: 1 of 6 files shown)


sbryngelson:
Thanks for the thorough review. Dispositions:

**2 (DRY stoichiometry walk) — done** (0226609). Extracted `_species_reaction_stoich(sol, species)`; both creation/destruction now share it. Verified output-preserving: regenerated Fortran before/after the refactor is byte-identical.

**4b (docstring) — done** (0226609). Dropped the "(production)" parenthetical so it no longer reads against `production_rate_expr` (the net rate).

**3 (convenience wrappers double-compute) — done** (aeb9d0a). Added a combined `get_creation_destruction_rates` across all three backends: one concentration + fwd + rev pass fills both `cdot` and `ddot` (Fortran out-args; Python/C++ return a pair). This is exactly the α-QSS hot-path need — it halves the rate-of-progress evaluations per sub-step (predictor+corrector went 4→2 passes in the downstream integrator). Added an identity assertion (combined == separate) to `test_kinetics`; validated combined == separate to 0.0 and == Cantera to ~8e-18. The C++ `std::pair` return binds via the existing `DEF_STATIC_METHOD` macro (`pybind11/stl.h` already included).

**1 (delegate `rate_of_progress_expr` to fwd − rev) — respectfully holding off.** `rate_of_progress_expr` builds the *factored* `k_fwd*(r_fwd − exp(log_keq)*r_rev)`, whereas fwd−rev is the *distributed* `k_fwd*r_fwd − k_fwd*exp(log_keq)*r_rev`. These are algebraically equal but not FP-identical, and pymbolic emits them as different code — so the refactor would change generated `get_net_production_rates` by a few ULP for every mechanism, turning an additive PR into one that perturbs every existing consumer's net-rate output (and downstream goldens). The divergence risk you flag is already caught by the `net == creation − destruction` machine-precision test across h2o2+gri30, so I'd rather keep the two halves explicit and lean on that guard. Happy to revisit if you'd prefer the single-source-of-truth tradeoff.

**4a (indices_fwd/indices_rev naming) / 4c (double `* ones`)** — noted; leaving 4a as a follow-up to preserve the additive diff, and 4c is intentional-by-convention as you noted.

PR #118, opened 2026-09-18
Generate code for heterogeneous surface kinetics

sbryngelson:
Adds code generation for heterogeneous (surface) kinetics from a `ct.Interface`, across all three backends.

## Why

Pyrometheus generates gas-phase thermochemistry in one call, but has nothing for surface chemistry, so anyone doing catalysis, ablation or char conversion has had to write the generator themselves. MFC did that (MFlowCode/MFC#1821): about 400 lines of Fortran assembled by string concatenation in its build toolchain, including its own NASA polynomial evaluation for non-gas species.

Two bugs in that copy were caught in review, both of the kind this PR avoids:

- a `hasattr(rate, "pre_exponential_factor")` guard accepted sticking-coefficient rates and emitted them as plain Arrhenius. Cantera exposes that attribute on every rate class, so the check passes, but for a sticking reaction those parameters are a dimensionless probability, not a rate coefficient.
- coverage-dependent rates lost their coverage terms.

## API

Each backend gains `generate_surface`, mirroring `generate`. The surface module holds only the surface mechanism and refers to the gas one; merging them would put two copies of the same NASA polynomials in one build.

```python
# gas, unchanged
FortranCodeGenerator.generate("thermochem", sol)
# surface
FortranCodeGenerator.generate_surface("surface_thermochem", interface,
                                      gas_module_name="thermochem")
```

| backend | surface artifact | refers to gas via |
|---|---|---|
| Python | `SurfaceThermochemistry` class | `gas` constructor argument |
| Fortran | `surface_thermochem` module | `use thermochem` |
| C++ | `SurfaceThermochemistry` struct | `GasT` template parameter |

Routine names agree across all three. Fortran needs the `surface_` prefix on the four the gas phase also defines, since a caller that `use`s both modules would otherwise have two `get_net_production_rates` in scope, so the same convention is used everywhere.

## What it handles

Three rate forms:

- interface-Arrhenius: the modified-Arrhenius form.
- sticking-Arrhenius: `k = γ/Γ₀ⁿ · √(RT/2πW)`, with the Motz-Wise form `γ/(1−γ/2)` when the mechanism asks for it.
- coverage dependence: `10^(aθ) · θ^m · exp(−Eθ/RT)` per dependent species.

Reversible surface reactions go through equilibrium constants, where the pressure-like factor is per phase: `(p₀/RT)^Δn_gas` for the gas, `Γ₀^Δn_surface` for the sites.

## Verification

`ptcombust.yaml` carries every form at once: 19 interface-Arrhenius, 5 sticking, 2 coverage-dependent, 3 reversible. The coverages used in the tests are non-uniform, since at uniform coverage a coverage-dependent rate is indistinguishable from a plain one.

Worst relative error against Cantera, over all 43 species of all three phases:

| | Python | Fortran (gfortran) | C++ (g++ -std=c++20) |
|---|---|---|---|
| site concentrations | 2.1e-16 | 2.1e-16 | 2.1e-16 |
| fwd rate coefficients | 5.6e-15 | 5.6e-15 | 5.6e-15 |
| rates of progress | 1.5e-14 | 1.5e-14 | 1.5e-14 |
| net production rates | 6.7e-15 | 6.7e-15 | 6.7e-15 |

The Fortran and C++ figures come from compiled binaries, not rendered source. The three backends also agree with each other to the same tolerance. That cross-check caught a real error: the kinetics ordering puts the interface's own species before the adjacent phases', and assembling the concentration vector gas-first gives a 3e+12 relative error.

## Tests and examples

Five tests in `test/test.py`, covering the expression layer, the generated Python class end to end, and that all three backends render (Mako reports template errors only at render time).

Two examples: `examples/surface_kinetics.py`, catalytic methane oxidation on platinum, and `examples/char_burnout.py`, char gasification with a bulk phase.

`flake8 --exclude examples` is clean. Note it only picks up the project's config through `Flake8-pyproject`; without it `max-line-length = 85` is ignored.

## Not included

Bulk-phase deposition and etch rates, and surface coverage ODE integration (Cantera's `advanceCoverages`). Neither is needed to couple a surface mechanism to a flow solver.


sbryngelson:
Context for reviewers on where this came from, and who is waiting on it.

MFC hand-wrote this generator in its own build toolchain because pyrometheus had no surface support — ~400 lines of Fortran built by string concatenation, including a re-implementation of NASA polynomial evaluation for non-gas species. Tracked downstream as MFlowCode/MFC#1891, under MFlowCode/MFC#1893.

That duplicate produced two silent bugs before review caught them, both of which this PR handles rather than reproduces:

- a `hasattr(rate, "pre_exponential_factor")` guard accepted **sticking-coefficient** rates and emitted them as plain Arrhenius. Cantera exposes that attribute on every rate class, so the probe passes — but for a sticking reaction those parameters are a dimensionless probability, not a rate coefficient.
- **coverage-dependent** rates were silently stripped of their coverage terms.

So the downstream swap deletes code and lifts restrictions at the same time.

The carbon mechanism added here as a second fixture is MFC's own, which is why it exercises a bulk phase, explicit reaction orders and `ConstantCp` thermo — none of which `ptcombust` has. Adding it is also what surfaced #119, fixed in #120.

rocfire11:
I agree that surface reactions really belonged in pyrometheus from the beginning, and more than grateful for the work you have done to move it out of simulations (including fixing the bugs) and into pyro. Please note the surface yaml file was created by hand because it uses a published global mechanism, that was not generated by gri or any other cantera mechanism.

README.rst
```diff
@@ -50,6 +51,21 @@ or via `Spack <https://spack.io>`__ (``spack install py-pyrometheus``)::
...
+      -s, --surface         Generate heterogeneous (surface) kinetics for an
+                            interface phase, rather than gas-phase
+                            thermochemistry. The adjacent phases are read from the
+                            mechanism.
+      --gas-name GAS_NAME   With --surface: name of the separately generated gas-
+                            phase module or header the surface code refers to
+                            (default: thermochem).
+
+A surface mechanism needs both: generate the gas phase first, then the interface,
+pointing the latter at the former::
```
ecisneros8: This idea can be conveyed in a more elegant way. The sentence degrades in quality.

pyrometheus/codegen/python.py
```diff
@@ -644,6 +644,248 @@ def get_species_mass_diffusivities_mixavg(self, pressure, temperature,
...
+surface_code_tpl = Template(
+    """\"""
+.. autoclass:: SurfaceKinetics
+\"""
+
+
+import numpy as np
+
+
+class SurfaceKinetics:
```
ecisneros8: I would rename this class. It computes thermodynamics too. 
In fact, it is making me rethink how we ought to name the gas-phase `Thermochemistry` class as well.  
Whereas this class could be `InterfaceThermochemistry` or `SurfaceThermochemistry`, the existing `Thermochemistry class could be `GasPhaseThermochemistry`.

pyrometheus/codegen/python.py
```diff
@@ -644,6 +644,248 @@ def get_species_mass_diffusivities_mixavg(self, pressure, temperature,
...
+        self.one_atm = ${repr(float(ct.one_atm))}
+
+        self.species_names = ${repr([interface.species_name(i)
+            for i in range(interface.n_species)])}
+        self.species_indices = dict([(name, i)
+            for i, name in enumerate(self.species_names)])
+        self.total_species_names = ${repr([interface.kinetics_species_name(k)
+            for k in range(interface.n_total_species)])}
+
+        # Where each phase's species start in the coupled ordering. A caller
```
ecisneros8: "coupled"?

pyrometheus/codegen/python.py
```diff
@@ -644,6 +644,248 @@ def get_species_mass_diffusivities_mixavg(self, pressure, temperature,
...
+    .. attribute:: species_indices
+
+    .. automethod:: get_site_concentrations
+    .. automethod:: get_surface_fwd_rate_coefficients
+    .. automethod:: get_surface_rates_of_progress
+    .. automethod:: get_surface_net_production_rates
+    .. automethod:: get_surface_net_heat_release_rate
+    \"""
+
+    def __init__(self, gas, pyro_np=np):
```
ecisneros8: Type-hint gas. Is it a Pyro-generated `Thermochemistry` object?

ecisneros8 (commented):
This is a valuable extension and overall great work. I want to read through it in more detail before merging, but my time has so far been limited. I left just a few comments that seek to clarify how this extension plans to handle the different phases. I'm hoping to come back to this PR and leave more comments later today.

pyrometheus/codegen/python.py
sbryngelson: Renamed to `SurfaceThermochemistry` in 8dd57a3. You are right that it computes thermodynamics: standard-state enthalpies, entropies and Gibbs functions for both the surface and bulk phases.

I went with `SurfaceThermochemistry` over `InterfaceThermochemistry` because the Fortran backend already emits a `surface_thermochem` module and every routine carries a `get_surface_` prefix, so it matches the rest of the generated code. Happy to switch if you would rather key the name to Cantera's `Interface`.

`get_surface_kinetics_class` follows to `get_surface_thermochem_class`.

On `GasPhaseThermochemistry`: I left it, and I should have checked before saying it breaks existing callers rather than asserting it. It does, and here is the reason. The Python backend was ignoring the `name` argument and always emitting `class Thermochemistry`, so every Python user has that exact name and could not have chosen otherwise. That part is now fixed in this PR, so callers can pick their own name, but the default is still `Thermochemistry` and renaming it would change what existing generate-to-a-file-and-import users get.

So it is a one-line change whose entire cost is downstream, which makes it your call rather than something to slip in here. Say the word and I will do it.

pyrometheus/codegen/python.py
sbryngelson: Dropped in b062b00. Everything is "total" now.

The class was using two words for one concept: `num_total_species` and `total_species_names`, which follow Cantera's `n_total_species`, alongside `coupled_*_offset`, `get_coupled_enthalpies_rt` and `num_coupled_species`. So `coupled_surface_offset` is now `total_surface_offset`, `get_coupled_enthalpies_rt` is `get_total_enthalpies_rt`, and the same across all three backends.

If you meant something else by it, say so and I will redo it.

pyrometheus/codegen/python.py
sbryngelson: Done in 8dd57a3, and my first answer here was wrong about why it could not be.

To answer the question: yes, it is a Pyro-generated gas object, whatever `get_thermochem_class` returned for the adjacent phase.

I said an annotation was impossible because the gas class had no fixed name to refer to. Checking that properly turned up two things. The Python backend was ignoring the `name` it was handed and always emitting `class Thermochemistry`, so `compile_class` could only ever look that one name up and no caller could pick another, unlike Fortran and C++ which both honor it. And `generate_surface` had no Python counterpart to Fortran's `gas_module_name` or C++'s `gas_header_name`, so there was genuinely nothing to name, but that was a gap in this PR rather than a fact about Python.

Both are fixed. `generate_surface` now takes `gas_module_name` and `gas_class_name`, and the generated class reads:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from thermochem import Thermochemistry


class SurfaceThermochemistry:
    def __init__(self, gas: "Thermochemistry", pyro_np=np):
```

Behind `TYPE_CHECKING` because the class takes a gas object at run time and has to stay importable when that module is not on the path, which is how `get_surface_thermochem_class` uses it. The generated code is still clean under the repo's generated-code flake8 job.

README.rst
sbryngelson: Rewritten in b062b00:

> A surface mechanism takes two passes. Generate the gas phase, then generate the interface against it::

pyrometheus/codegen/python.py
sbryngelson: Correcting myself again on the cost of `GasPhaseThermochemistry`. I said it breaks existing callers. I went and measured it, and that is mostly not true.

Every consumer I can find gets the class through `get_thermochem_class`, which returns the class object, so the name never appears in their source:

| consumer | how it gets the class | affected |
|---|---|---|
| MFC | `FortranCodeGenerator().generate("m_thermochem", sol)` | no, Fortran already honored the name |
| mirgecom | `get_thermochem_class(sol)` | no |
| drivers_flame1d | `get_thermochem_class(sol)` | no |
| this repo's tests | `get_thermochem_class`, or an explicit `-n` | no |

I also searched for anyone importing the generated class by name and found nothing pyrometheus-related. The breaking case is someone who generates to a file and then does `from thermochem import Thermochemistry`, and I cannot find one.

Now that the backend honors the `name` it is given, the rename is three lines:

```
python.py:1022    name = "Thermochemistry"    -> "GasPhaseThermochemistry"
python.py:953     gas_class_name default      -> "GasPhaseThermochemistry"
test/test.py:1154 the CLI test's -n argument
```

What changes for users is `__name__` in reprs and tracebacks, and the text of generated files for anyone using the default. If you want it strictly non-breaking, one alias line in the generated module would do it, at the cost of a line in every generated file.

Say the word and I will do it in this PR, or leave it for its own if you would rather not mix it in.

pyrometheus/codegen/python.py
sbryngelson: One thing I should flag about this, since I picked an answer without saying there was a choice.

The class was using two words for one concept, so I settled on one. But there were two Cantera-derived candidates and I took the first:

- `total_*`, from `n_total_species`, which the class already mirrors in `num_total_species` and `total_species_names`. This is what I shipped.
- `kinetics_*`, from `kinetics_species_name`, which is arguably the more direct Cantera term for this particular ordering.

So `total_surface_offset` and `get_total_enthalpies_rt` could equally be `kinetics_surface_offset` and `get_kinetics_enthalpies_rt`. Either is a small change at this point.

If you meant something else by the original comment, for instance that the offsets should not be on the class at all, that is a different change and I would rather hear it than guess twice.

pyrometheus/codegen/python.py
```diff
@@ -644,6 +644,257 @@ def get_species_mass_diffusivities_mixavg(self, pressure, temperature,
...
+        self.total_species_names = ${repr([interface.kinetics_species_name(k)
+            for k in range(interface.n_total_species)])}
+
+        # Where each phase's species start in the total-species ordering.
+        # Which phase follows the interface's own is set by the mechanism.
+        %for kind, start, stop, src in phase_blocks:
+        self.total_${kind}_offset = ${start}
+        %endfor
+
+    def _pyro_make_array(self, res_list):
```
ecisneros8: I don't see a need to re-implement this routine. It is identical to the one provided by `Thermochemistry`.  At some point, perhaps not now, there should be an abstract class from which individual phases derive.

pyrometheus/codegen/python.py
```diff
@@ -672,6 +923,7 @@ def generate(name: str, sol: ct.Solution,
 
         return code_tpl.render(
             ct=ct,
+            name=name,
```
ecisneros8, line 947: `name` is too vague. Substitute with `gen_cls_name` or even `generated_cls_name`.

pyrometheus/codegen/python.py
```diff
@@ -695,6 +947,73 @@ def compile_class(name: str, source: str):
 
         return exec_dict[name]
 
+    @staticmethod
+    def generate_surface(name: str, interface: ct.Interface,
```
ecisneros8: This routine does not generate a surface. Replace name with `generate_surface_thermochem`.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+
+    if n_gas_phases > 1:
+        raise ValueError(
+            f"interface '{interface.name}' couples {n_gas_phases} gas phases; the "
+            "generated code takes a single gas-phase class")
+
+    return blocks
+
+
+def _surface_needs_gas_standard_concentration(interface: ct.Interface):
```
ecisneros8, line 792: This routine does not belong here. It does not return a symbolic expression. This helper can be embedded as a closure in the routines that invoke it within the codegen module.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    couples, but they do not all come from the same place: the surface species and
+    any bulk species are generated with the surface code, while the gas-phase ones
+    come from the separately generated gas class. This returns one block per phase,
+    in kinetics order, as ``(kind, start, stop, source_start)``: *kind* is
+    ``"surface"``, ``"gas"`` or ``"bulk"``, ``start:stop`` the block's span in the
+    kinetics ordering, and *source_start* its offset within its own source array,
+    which differs from *start* only for bulk phases, since they share one array
+    while not necessarily being adjacent in the kinetics ordering.
+
+    Raises if the interface couples more than one gas phase, which the generated
```
ecisneros8, line 765: Fine by me for now, but we should work on this problem. It would be a good idea to open a corresponding Issue.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+            for k in range(interface.n_total_species)]
+
+
+# Length dimension of the standard concentration: kmol/m^3 in a volume,
+# kmol/m^2 on a surface, dimensionless in a bulk solid. Classifying on the
+# thermo model would mean enumerating every model name Cantera has.
+_PHASE_KIND_BY_LENGTH_DIMENSION = {-3.0: "gas", -2.0: "surface", 0.0: "bulk"}
+
+
+def surface_phase_kind(phase):
```
ecisneros8, line 722: I'm not sure this routine belongs in the `chem_expr` module. Again, this module constructs symbolic expressions, to be mapped to code when rendering the template. If this routine is called only inside `chem_expr`, then it can stay here, with a leading underscore.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    units = phase.standard_concentration_units
+    try:
+        return _PHASE_KIND_BY_LENGTH_DIMENSION[units.dimensions["length"]]
+    except KeyError:
+        raise ValueError(
+            f"phase '{phase.name}' has standard concentration units '{units}', "
+            "which heterogeneous kinetics does not handle") from None
+
+
+def _surface_kinetics_phases(interface: ct.Interface):
```
ecisneros8, line 736: Another routine that might not belong here. How does it serve the construction of symbolic expressions? Is it called elsewhere?

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    mechanism whose reversible reactions are all of that kind never references
+    the factor.
+    """
+    return any(
+        react.reversible
+        and _surface_standard_concentrations(interface, i)[0]
+        for i, react in enumerate(interface.reactions()))
+
+
+def surface_bulk_species(interface: ct.Interface):
```
ecisneros8: What does this routine return? Type-hint its output.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    return [sp
+            for phase, _offset in _surface_kinetics_phases(interface)
+            if phase is not interface and surface_phase_kind(phase) != "gas"
+            for sp in phase.species()]
+
+
+def _surface_standard_concentrations(interface: ct.Interface, reaction_index):
+    """Standard-concentration contribution of a reaction, split by phase kind.
+
+    :returns: the net change in moles of gas-phase species, whose standard
```
ecisneros8: The return type is not mentioned in this docstring.

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    Their thermodynamics is generated with the surface code: unlike the gas phase,
+    a bulk phase has no Pyrometheus class of its own to defer to.
+    """
+    return [sp
+            for phase, _offset in _surface_kinetics_phases(interface)
+            if phase is not interface and surface_phase_kind(phase) != "gas"
+            for sp in phase.species()]
+
+
+def _surface_standard_concentrations(interface: ct.Interface, reaction_index):
```
ecisneros8: Not even sure what this routine does. The function name and docstring seem at odds. Does this routine produce expressions for the standard concentrations (per the name), or a net molar change in concentrations (as per the docstring)?

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+    """Standard-concentration contribution of a reaction, split by phase kind.
+
+    :returns: the net change in moles of gas-phase species, whose standard
+        concentration is temperature-dependent and so stays symbolic, and the log
+        of the surface contribution, which is a number.
+
+    Bulk species have unit activity and drop out; lumping them in with the gas
+    species would leave a spurious factor of :math:`(p_0/RT)^{\\Delta n_b}`, which
+    is two orders of magnitude per mole of bulk at combustion temperatures.
```
ecisneros8: By "orders of magnitude", do you mean a factor of 100?

pyrometheus/chem_expr.py
```diff
@@ -568,4 +588,345 @@ def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
...
+
+
+def _surface_standard_concentrations(interface: ct.Interface, reaction_index):
+    """Standard-concentration contribution of a reaction, split by phase kind.
+
+    :returns: the net change in moles of gas-phase species, whose standard
+        concentration is temperature-dependent and so stays symbolic, and the log
+        of the surface contribution, which is a number.
+
+    Bulk species have unit activity and drop out; lumping them in with the gas
```
ecisneros8: I've found that this kind of comment sow more confusion than they attempt to dispel. This kind of information should go into the documentation.

ecisneros8 (commented):
A few more comments.

pyrometheus/codegen/python.py
sbryngelson: Removed in 61b7744. The surface class holds the gas object already, so the twelve call sites now go through `self.gas._pyro_make_array`.

Agreed on the abstract base class. That is the same shape as the naming question in your other comment, and I have not tried it here.

pyrometheus/codegen/python.py
sbryngelson: Renamed to `generate_surface_thermochem` in 61b7744, in all three backends and the CLI.

pyrometheus/chem_expr.py
sbryngelson: Type-hinted in 61b7744: `-> list`, with the docstring saying it is a list of `cantera.Species`.

pyrometheus/chem_expr.py
sbryngelson: Added in 61b7744. It returns a `(float, float)` pair, and the docstring now says which is which.

pyrometheus/chem_expr.py
sbryngelson: You are right, they were at odds. The name described the quantity it helps build, not the one it returns.

Renamed to `_surface_standard_concentration_exponents` in 61b7744. What it returns is the exponent on `p0/RT` for the gas side and the log of the surface contribution, and the docstring now says so with the types.

pyrometheus/chem_expr.py
sbryngelson: Fair. It is `RT/p0` per mole of solid: 123 at 1500 K, 164 at 2000 K, 205 at 2500 K. Not exactly 100, so the phrase was doing no work.

61b7744 puts those numbers in the model documentation and drops the vague wording.

pyrometheus/chem_expr.py
sbryngelson: Moved to the model documentation in 61b7744, under the equilibrium-constant section, and the comment is gone.

pyrometheus/chem_expr.py
sbryngelson: Opened as #121, and the docstring points at it.

It covers what would have to change: each backend names the gas phase through a single handle, the `gas` argument, `use <module>`, or the `GasT` parameter, and a second gas phase needs all three to become collections. The kinetics ordering already carries the block-per-phase information.

pyrometheus/chem_expr.py
sbryngelson: `_surface_kinetics_phases` is called only inside `chem_expr`, zero references anywhere else, and it already carries the underscore. By the criterion in your other comment it can stay, so I have left it.

On how it serves the expressions: `surface_rate_of_progress_expr` and the equilibrium-constant expressions need to know which slice of the kinetics ordering belongs to which phase, because the activity concentration differs by phase kind. It answers that. If you would rather that mapping lived in the codegen module, it is a contained move and I am happy to do it.

pyrometheus/chem_expr.py
sbryngelson: `surface_phase_kind` is called outside `chem_expr`, but only by the test suite, `test/test.py:772`, which uses it to decide whether a phase is bulk when setting up a state. Nothing in `codegen` calls it.

So it does not meet the "called only inside `chem_expr`" bar, but only because of a test. I can underscore it and have the test classify phases itself, or leave it public. Which would you prefer?

On whether it belongs here at all: it maps a phase to gas, surface or bulk by the length dimension of its standard concentration, which is the thing that decides what a species contributes to a rate of progress. That is why it sits next to the expression builders, but I would not argue it is symbolic.

pyrometheus/chem_expr.py
sbryngelson: Happy to take it out of `chem_expr`. One wrinkle with the closure form: `_surface_needs_gas_standard_concentration` is now called by all three backends, `python.py`, `fortran.py` and `cpp.py`, since each gates the `p0/RT` prologue on it. A closure in each render routine writes it three times.

Would you rather it moved to a shared spot under `codegen/`, or is the triplication acceptable to you? I have left it where it is pending your answer rather than pick for you.

pyrometheus/codegen/python.py
sbryngelson: `name` is the first positional parameter of `generate` and `generate_surface_thermochem` on all three backends, so renaming it in Python alone would break the symmetry between them.

Do you want `gen_cls_name` everywhere? For Fortran it names a module and for C++ a struct, so "cls" does not quite fit either. `gen_name` or `generated_name` would cover all three, or the backends could each use the word that fits, at the cost of a common signature.

Tell me which and I will do it in one pass. Left alone for now.

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+
+Heterogeneous Kinetics
+----------------------
+
+Pyrometheus also generates code for reactions that take place on a surface rather
+than in the bulk gas. These are described by a Cantera *interface* phase, which
+couples the surface sites to the phases that react at them: a gas phase and,
+for mechanisms such as char gasification, one or more bulk (solid) phases.
+
+Throughout, :math:`N_{s}` is the number of surface species, :math:`N_{c}` the
```
ecisneros8, line 126: Throughout what?

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+number of species the interface couples in total, and :math:`M_{s}` the number of
```
ecisneros8, line 127: "number of species the interface couples in total" -> "number of species coupled by the interface"

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+heterogeneous reactions. Concentrations and production rates over all coupled
+phases are ordered as Cantera orders them for kinetics: the interface's own
+species first, then those of each adjacent phase in turn. This is *not* the
```
ecisneros8, line 130: Drop the "This is *not* ...". Unnecessary, given that the order has already been explained.

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+the site density of the interface (in :math:`\mathrm{kmol/m^{2}}`), its
+concentration is
+
+.. math::
+
+   \label{eq:site_concentration}
+     [\mathcal{S}_{k}] = \frac{ \theta_{k}\Gamma_{0} }{ \sigma_{k} },\qquad
+     k = 1,\dots,N_{s}.
+
+The factor :math:`\sigma_{k}` is easy to overlook, since the great majority of
```
ecisneros8, line 152: Super confusing statement. I would take it out. One could say :math:`\sigma_{k}` often takes a value of 1.

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+concentration, and each of the :math:`n_{j}` sites the reaction consumes divides
+by the site density, so
+
+.. math::
+
+   \label{eq:sticking}
+     k_{j}(T) = \frac{ \gamma_{j}(T) }{ \Gamma_{0}^{n_{j}} }
+     \sqrt{ \frac{RT}{2\pi W_{m}} }.
+
+Reading :math:`A_{j}`, :math:`b_{j}` and :math:`\theta_{a,j}` as though they were
```
ecisneros8, line 227: This kind of statement has no place in our documentation. It is confusing. We write plainly what we do, and why.

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
 Species Thermodynamics
 ~~~~~~~~~~~~~~~~~~~~~~
 
+.. _subsec:surface:
+
+Heterogeneous Kinetics
+----------------------
+
+Pyrometheus also generates code for reactions that take place on a surface rather
+than in the bulk gas. These are described by a Cantera *interface* phase, which
```
ecisneros8, line 122: Subsequent text uses "bulk" for the solid. Thus, "bulk gas" is confusing. Ensure consistency. I do not know what nomenclature Cantera uses, but it can serve as a guide. In general, I think the word "bulk" is used quite arbitrarily across this PR. It should be clearly and unambiguously defined.

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+
+The factor :math:`\sigma_{k}` is easy to overlook, since the great majority of
+mechanisms give every surface species a single site.
+
+Because the species that meet at an interface live in phases of different
+dimensionality, the quantity that enters a rate of progress is not the molar
+concentration for all of them but the *activity concentration*: the molar
+concentration for a gas species (in :math:`\mathrm{kmol/m^{3}}`), the site
+concentration `[site_concentration] <#site_concentration>`__ for a surface
+species (in :math:`\mathrm{kmol/m^{2}}`), and the activity for a bulk species,
```
ecisneros8, line 160: What is a bulk species?

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+the forward direction alone, so the reverse product always uses
+:math:`\nu_{kj}^{\prime\prime}`. The production rates then follow exactly as in
+the gas phase,
+
+.. math::
+
+   \dot{\omega}_{i} = \sum_{j = 1}^{M_{s}}\nu_{ij}R_{j},\qquad i = 1,\dots,N_{c},
+
+and cover every phase the interface couples: gas species consumed or released at
+the wall, surface species whose coverages evolve, and bulk species consumed as
```
ecisneros8, line 195: What is a bulk species?

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+identical, the second following from the first by
+`[production_rates] <#production_rates>`__; the first is what the generated code
+evaluates, because it needs the stoichiometry only once, inside
+:math:`\dot{\omega}`.
+
+The sum runs over every species the interface couples, so a reaction that
+consumes solid carries the enthalpy of the solid it consumes. Like the Gibbs
+functions of `[surface_equil_constants] <#surface_equil_constants>`__, the gas-phase
+enthalpies come from the separately generated gas-phase code and the surface and
+bulk ones are generated alongside the surface mechanism.
```
ecisneros8, line 271: What is the bulk?

doc/model.rst
```diff
@@ -113,6 +113,205 @@ the species enthalpies and entropies.
...
+   c^{0}_{i} = \begin{cases}
+     p_{0}/RT, & \text{$i$ a gas species,} \\
+     \Gamma_{0}/\sigma_{i}, & \text{$i$ a surface species,} \\
+     1, & \text{$i$ a bulk species.}
+   \end{cases}
+
+A reaction that changes the moles of gas thus carries
+:math:`(p_{0}/RT)^{\Delta n_{g,j}}` as it does in the gas phase, one that changes
+the number of occupied sites carries the corresponding power of the site density,
+and a bulk species contributes nothing at all. Treating a bulk species as though
```
ecisneros8, line 304: Without clearly defining the bulk, this discussion is confusing.

examples/char_burnout.py
```diff
@@ -0,0 +1,172 @@
+"""Burnout of a char particle.
+
+A global char mechanism: five irreversible gasification channels on solid
+carbon, one site species, and three phases.
+
+    surface   C-site                       total_surface_offset
+    gas       CO CO2 H2 H2O O2 N2 ...      total_gas_offset
+    bulk      C(gr)                        total_bulk_offset
```
ecisneros8, line 8: So is bulk the solid? Why not just use solid?

pyrometheus/codegen/fortran.py
```diff
@@ -1015,6 +1015,314 @@ def map_if(self, expr, enclosing_prec):
...
+        ${real_type}, intent(in)  :: temperature
+        ${real_type}, intent(out) :: s0_r(num_surface_species)
+
+        %for k, sp in enumerate(interface.species()):
+        s0_r(${k+1}) = ${cgm(ce.poly_to_entropy_expr(
+            sp.thermo, "temperature"))}
+        %endfor
+
+    end subroutine get_surface_entropies_r
+%if bulk_species:
```
ecisneros8, line 1091: Another instance of "bulk". What is it?

pyrometheus/codegen/fortran.py
```diff
@@ -1015,6 +1015,314 @@ def map_if(self, expr, enclosing_prec):
...
+
+    ! Standard-state Gibbs energy over RT of every bulk
+    ! species. A bulk phase has no Pyrometheus module of
```
ecisneros8, line 1094: So, is this the defining property of a bulk phase? 
It is somewhat self-defeating though because _this_ very module defines the thermo for so-called "bulk" species.

pyrometheus/codegen/python.py
ecisneros8: `gen_cls_name` for Python, `gen_module_name` for Fortran, `gen_struct_name` for C++.

pyrometheus/chem_expr.py
ecisneros8: I'll allow call exceptions in the test suite. However, because it is only a helper and does not return a symbolic expression, its name should be preceded by an underscore -> `_surface_phase_kind.

pyrometheus/chem_expr.py
ecisneros8: I would much rather it live in `codegen/` for now. Seek for an unassuming spot there. This kind of juggling is what the bandit aims to address, so these changes will be superseded in Pyro 2.0.

ecisneros8 (commented):
Most of my earlier comments are now resolved. My biggest concern is the use of "bulk" as a phase category. It is used in the docs to refer to both the gas and the solid. Then, a docstring seems to indicate it corresponds to species with no codegen'd module, followed by codegen for their proerties. At any, I would like to see a clear definition of phases, and a more judicious choice of labels.

PR #120, opened 2026-09-18
Generate working code for mechanisms with no reactions

### LINKED ISSUE #119: Generating from a reaction-free mechanism emits unused locals (F841)
Generating a Python module from a mechanism with **no reactions** produces code that fails this repo's own generated-code lint.

## Reproduce

Any reaction-free mechanism does it. Using the surface fixture's gas phase (11 species, 0 reactions):

```python
import cantera as ct
from pyrometheus.codegen.python import PythonCodeGenerator

sol = ct.Solution("test/surface_mechs/carbon_surface.yaml", "gas")
open("out.py", "w").write(PythonCodeGenerator.generate("pyrometheus", sol))
```

```
$ python3 -m flake8 --extend-ignore E501,Q000 out.py
out.py:329:9: F841 local variable 'c0' is assigned to but never used
out.py:331:9: F841 local variable 'g0_rt' is assigned to but never used
out.py:359:9: F841 local variable 'ones' is assigned to but never used
out.py:366:9: F841 local variable 'k_fwd' is assigned to but never used
out.py:367:9: F841 local variable 'log_k_eq' is assigned to but never used
out.py:390:9: F841 local variable 'k_fwd' is assigned to but never used
out.py:395:9: F841 local variable 'k_fwd' is assigned to but never used
out.py:396:9: F841 local variable 'log_k_eq' is assigned to but never used
```

## Cause

Routines like `get_equilibrium_constants` and `get_net_rates_of_progress` declare their working variables unconditionally and then emit one term per reaction. With zero reactions the declarations remain and nothing consumes them.

## Why it matters

The code still *runs* — these are unused locals, not errors — so the impact is limited to lint. But it is the project's own `Flake8 on Generated Code` job that fails, so a user who adds a reaction-free mechanism to `test/mechs/` gets a red CI leg with no obvious cause. That is how this was found (#118).

Inert or transport-only mixtures are the realistic case: a mechanism used purely for thermo and diffusion, with kinetics supplied elsewhere.

## Not urgent

Nothing generated today is affected — every mechanism in `test/mechs/` has reactions. Filing so it is not lost.

sbryngelson:
Closes #119.

## The bug

#119 reported unused locals (`F841`) in code generated from a reaction-free mechanism. The generated code also crashes when called:

```python
sol = ct.Solution("test/mechs/inert.yaml")     # 3 species, 0 reactions
thermochem = PythonCodeGenerator.get_thermochem_class(sol)()
thermochem.get_net_production_rates(rho, T, y)
```
```
IndexError: index 0 is out of bounds for axis 0 with size 0
```

`get_net_production_rates`, `get_creation_rates` and `get_destruction_rates` all raise. The module imports fine and every other routine works, so nothing surfaced it until the code was called.

## Cause

Two prologues that assume at least one reaction exists:

1. The array shape. Vectorized evaluation needs a broadcasting `1` shaped like the data, taken from `r_net[0]` / `r_fwd[0]`. With no reactions those vectors are empty and the index is invalid.
2. Unused fetches. `c0`, `g0_rt`, `k_fwd`, `log_k_eq`, `r_net`, `r_fwd` and `r_rev` are computed to feed per-reaction terms that are never emitted.

## Fix

The shape comes from `temperature` when there are no reactions, which is always in scope and correct for both scalar and array use. The prologues that exist only to feed reaction terms are emitted only when there are reactions.

Production, creation and destruction rates now come out as correctly shaped zeros and match Cantera.

## Fixture

`test/mechs/inert.yaml`, a species-and-thermo mixture with `reactions: none`, which covers transport-only and frozen-chemistry calculations. It sits in `test/mechs/` so the existing Flake8 on Generated Code job generates from it and the lint half of the bug cannot return.

## Verification

- The new test fails on `main` with the `IndexError` above and passes here.
- `flake8 --extend-ignore E501,Q000 test/mechs` after generating from every mechanism: 0 errors, against 11 for a reaction-free one before.
- Mechanisms with reactions are untouched: `sandiego` (24 reactions) agrees with Cantera to 3.3e-16, `uiuc` (3 reactions) exactly, as before.
- `flake8 --exclude examples`: clean.

## Scope

Python backend only, which is where both the crash and the lint job are. Generating `inert.yaml` through the Fortran and C++ backends gives zero-length arrays and no indexing of a rate vector, so neither has the equivalent problem.


sbryngelson:
Found while adding a second surface-mechanism fixture in #118: that mechanism's gas phase has species and thermo but no kinetics, and generating from it produced code that failed this repo's own `Flake8 on Generated Code` job — then, on investigation, turned out to raise `IndexError` when called.

Independent of #118 and branched from `main`, so it can land in either order. Filed as #119.

Downstream context, for anyone wondering why a reaction-free mechanism came up at all: MFlowCode/MFC#1891.

pyrometheus/chem_expr.py
```diff
@@ -182,6 +182,22 @@ def _zeros_like(argument):
     # FIXME: This mishandles NaNs.
     return 0 * argument
 
+
+def _ones_from(sol: ct.Solution, rates, fallback):
```
ecisneros8: Why not `_ones_like`?

pyrometheus/chem_expr.py
```diff
@@ -464,13 +480,13 @@ def rate_of_progress_expr(sol: ct.Solution, reaction_index, c,
...
 
-def production_rate_expr(sol: ct.Solution, species, r_net):
+def production_rate_expr(sol: ct.Solution, species, r_net, fallback=None):
     """
     :returns: Species production rate for species *species* in terms of
         the net reaction rates of progress *r_net* as a
         :class:`pymbolic.primitives.Expression`
     """
-    ones = _zeros_like(r_net[0]) + 1.0
+    ones = _ones_from(sol, r_net, fallback)
```
ecisneros8: I do not understand why we ought to concern ourselves with array sizes `chem_expr`. This module focuses strictly creating symbolic expressions. Array considerations can be handled in the template. AFAIK, there already are instances in the template that can serves as examples. The generated Thermochemistry class has a `_ones_like`-like function for this kind of work.

pyrometheus/chem_expr.py
```diff
@@ -539,27 +555,27 @@ def _species_reaction_stoich(sol: ct.Solution, species):
...
 
-def creation_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
+def creation_rate_expr(sol: ct.Solution, species, r_fwd, r_rev, fallback=None):
     """
     :returns: Species creation rate for *species*: created as a product by
         forward reactions and as a reactant by reverse reactions. Mirrors
         Cantera's ``creation_rates``.
     """
-    ones = _zeros_like(r_fwd[0]) + 1.0
+    ones = _ones_from(sol, r_fwd, fallback)
```
ecisneros8: As in the preceding comment, unclear that such concerns should be expressed in this module.

pyrometheus/chem_expr.py
```diff
@@ -539,27 +555,27 @@ def _species_reaction_stoich(sol: ct.Solution, species):
...
 
-def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev):
+def destruction_rate_expr(sol: ct.Solution, species, r_fwd, r_rev, fallback=None):
     """
     :returns: Species destruction rate for *species*: consumed as a reactant by
         forward reactions and as a product by reverse reactions. Mirrors
         Cantera's ``destruction_rates``. creation - destruction == net.
     """
-    ones = _zeros_like(r_fwd[0]) + 1.0
+    ones = _ones_from(sol, r_fwd, fallback)
```
ecisneros8: Same concern as in my other comments.

ecisneros8 (commented):
Thanks for working on this issue. It is a good catch.

pyrometheus/chem_expr.py
sbryngelson: Fixed in 72d7524. `chem_expr` is now identical to `main`.

You were right that this was already handled elsewhere. The template computes

```
ones = self._pyro_zeros_like(${shape_src('r_net')}) + 1.0
```

and multiplies each species expression by it, so the `fallback` I threaded into `production_rate_expr` was doing the same job twice.

It works with no help from `chem_expr` because with no reactions the species sum is empty, pymbolic folds it to `0.0`, and the expression never mentions `r_net`, so there is nothing to index. `shape_src` falls back to `temperature` and the generated body is `0.0 * ones` per species.

I checked this by generating every mechanism in `test/mechs` for all three backends both ways: the output is byte-identical, including `inert.yaml`. The zero-reaction test still passes and the generated code is still flake8-clean.

pyrometheus/chem_expr.py
sbryngelson: Gone entirely. `_ones_from` went with the `chem_expr` revert in 72d7524, so `_zeros_like` is again the only helper here.

pyrometheus/chem_expr.py
sbryngelson: Same fix. 72d7524 reverts `chem_expr` to `main`, so `creation_rate_expr` keeps its three-argument form.

pyrometheus/chem_expr.py
sbryngelson: Same fix. `destruction_rate_expr` is back to its `main` form in 72d7524.

pyrometheus/codegen/python.py
```diff
@@ -670,10 +708,16 @@ def generate(name: str, sol: ct.Solution,
...
 
+        def shape_src(rates):
+            # With no reactions there is no rates[0] to take an array shape from.
+            return f"{rates}[0]" if sol.n_reactions else "temperature"
+
         return code_tpl.render(
             ct=ct,
             sol=sol,
 
+            shape_src=shape_src,
```
ecisneros8: This solution looks over-engineered. Why can't modify the template so that the routines in question return an empty array if `sol.n_reactions == 0` through a `if ... elif ...` branch?

pyrometheus/codegen/python.py
```diff
@@ -515,48 +531,70 @@ def get_rev_rates_of_progress(self, temperature, concentrations):
...
                 Variable("r_fwd"), Variable("r_rev")))} <%
             %>* ones,
+            %else:
+            0.0 * ones,
+            %endif
             %endfor
             ])
 
     def get_creation_destruction_rates(self, rho, temperature, mass_fractions):
+        %if sol.n_reactions:
```
ecisneros8, line 569: Why are there no `elif` and `else` branches? This `if` is not encoding all cases. It is a missed opportunity to return an empty array if `sol.n_reactions == 0`.

sbryngelson: Restructured in bc1fd54. Each of the four per-species routines now branches once at the top, `%if sol.n_reactions:` with an `%else:`, instead of carrying the empty case through the normal path.

Not an empty array, though. These four return one entry per species, so for a three-species mechanism with no reactions the answer is three zeros, and Cantera agrees:

```
inert.yaml: 3 species, 0 reactions
  net_production_rates   cantera (3,)   ours (3,)
  creation_rates         cantera (3,)   ours (3,)
  destruction_rates      cantera (3,)   ours (3,)
  net_rates_of_progress  cantera (0,)   ours (0,)
  forward_rate_constants cantera (0,)   ours (0,)
```

The per-reaction routines are the ones where empty is right, and they already return empty. The `%else:` branch now reads

```
zero = self._pyro_zeros_like(temperature)
return self._pyro_make_array([zero] * 3)
```

pyrometheus/codegen/python.py
sbryngelson: `shape_src` is gone in bc1fd54.

It existed to pick the array-shape source for a single `ones` expression shared by both cases. With each routine branching explicitly, the empty case builds its own array from `temperature` and the helper has nothing left to do.

